PR #1, opened 2022-10-27
pkvm: x86: nested performance optimization: remote TLB flush

ZideChen0:


James-Zzz (commented):
For commit msg of "pkvm: x86: flush remote TLBs with range in mmu_notifier callbacks"

KVM could cache multiple shadow MMU roots and we can't pass down the
shadow root infomation with the existing tlb_remote_flush_with_range
callback. 

There is a typo: “infomation“ -> "information"

ZideChen0:
> 

Fixed

PR #8, opened 2022-12-02
Memory protection with legacy IOMMU in pass-through mode

semihalf-maluka-dmytro:
Add initial support for secure VM memory protection from DMA in legacy IOMMU mode. For now protection is ensured only for cases when the host uses vIOMMU in pass-through mode, i.e. when the host itself doesn't use any address translation for the given device. In this case in pKVM we enforce address translation and use the host EPT as the IOMMU page tables.

### DIFF
--- a/arch/x86/kvm/vmx/pkvm/hyp/iommu.c
+++ b/arch/x86/kvm/vmx/pkvm/hyp/iommu.c
@@ -348,10 +348,9 @@ static bool sync_shadow_context_entry(struct pgt_sync_data *sdata)
 	 * After the missing piece of shadowing paging structures of legacy mode is
 	 * added, we won't need this workaround anymore.)
 	 */
-	if ((sdata->guest_ptep && sdata->shadow_pa) ||
+	if (sdata->guest_ptep && (sdata->shadow_pa ||
 		(!ecap_smts(sdata->iommu_ecap) &&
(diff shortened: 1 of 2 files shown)


arch/x86/kvm/vmx/pkvm/hyp/iommu.c
```diff
@@ -321,41 +321,33 @@ static bool sync_shadow_context_entry(struct pgt_sync_data *sdata)
...
-		 (context_lm_get_tt(guest_ce) == CONTEXT_TT_PASS_THROUGH)))) {
+	if (sdata->guest_ptep && sdata->shadow_pa) {
 		tmp.hi = guest_ce->hi;
 		tmp.lo = sdata->shadow_pa | (guest_ce->lo & 0xfff);
 
-		/* Clear DTE to make sure device TLB is disabled for security */
-		context_clear_dte(&tmp);
+		if (ecap_smts(sdata->iommu_ecap))
+			/* Clear DTE to make sure device TLB is disabled for security */
+			context_clear_dte(&tmp);
```
JasonChenCJ, line 332: thanks for fixing this potential issue, maybe we should change the function name to context_sm_clear_dte().

JasonChenCJ (approved):

TinaZhangZW (approved):

TinaZhangZW (approved):

PR #15, opened 2022-12-22
Fix detection of IOMMU coherency in legacy mode

semihalf-maluka-dmytro:
The ECAP.SMPWC (scalable-mode page-walk coherency) capability bit is valid only in scalable mode. In legacy mode the coherency of IOMMU paging structures is reported via the ECAP.C (page-walk coherency) bit instead.

### DIFF
--- a/arch/x86/kvm/vmx/pkvm/pkvm_host.c
+++ b/arch/x86/kvm/vmx/pkvm/pkvm_host.c
@@ -169,8 +169,7 @@ static int check_and_init_iommu(struct pkvm_hyp *pkvm)
 		iounmap(addr);
 
 		/*
-		 * If pkvm IOMMU works in scalable mode, it needs to check SMPWC for
-		 * the coherency of the paging structure accessed through pasid table entry.
+		 * If pkvm IOMMU works in scalable mode, it requires to use nested translation.
 		 */
 		if (ecap_smts(ecap) && !ecap_nest(ecap)) {
 			pr_err("pkvm: drhd reg_base 0x%llx: nested translation not supported\n",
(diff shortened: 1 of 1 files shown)


arch/x86/kvm/vmx/pkvm/pkvm_host.c
```diff
@@ -75,6 +75,15 @@ static int check_pci_device_count(void)
...
 
+/*
+ * Check for the coherency of paging structures accessed through pasid table
+ * entries (in scalable mode) or context table entries (in legacy mode).
+ */
+static inline bool is_iommu_coherent(u64 ecap)
+{
+	return ecap_smts(ecap) ? !!ecap_smpwc(ecap) : !!ecap_coherent(ecap);
+}
+
```
JasonChenCJ, line 86: do we finally need make it as a public API, or just need use it in pkvm_host.c

semihalf-maluka-dmytro: Possibly we will still use it just in pkvm_host.c, since once we use it to initialize `pkvm->iommu_coherent`, we just use `pkvm->iommu_coherent` value.

OTOH this function is actually the same as `iommu_paging_structure_coherency()` in `drivers/iommu/intel/iommu.c`. This is just one more example of duplication of various Intel IOMMU related definitions in pKVM and in the host IOMMU driver (other examples are context and pasid entry related functions in `arch/x86/kvm/vmx/pkvm/hyp/iommu_internal.h`). Maybe finally we'll want to remove this duplication by unifying those definitions in a common header, e.g. in `include/linux/intel-iommu.h`.

JasonChenCJ: Looks good to me, yes, finally we shall unify with common header file.

PR #19, opened 2023-02-15
Flush cache for pgstate_pgt

cxdong:
Pgstate_pgt can be used as IOMMU page table when protected VM has passthrough device. If IOMMU is not coherent in this case, flush cache is needed when modify the pgstate_pgt.

### DIFF
--- a/arch/x86/kvm/vmx/pkvm/hyp/ept.h
+++ b/arch/x86/kvm/vmx/pkvm/hyp/ept.h
@@ -48,6 +48,7 @@ void pkvm_shadow_clear_suppress_ve(struct kvm_vcpu *vcpu, unsigned long gfn);
 
 int pkvm_pgstate_pgt_init(struct pkvm_shadow_vm *vm);
 void pkvm_pgstate_pgt_deinit(struct pkvm_shadow_vm *vm);
+void pkvm_pgstate_pgt_update_coherency(struct pkvm_pgtable *pgt, bool coherent);
 bool is_pgt_ops_ept(struct pkvm_pgtable *pgt);
 
 static inline bool is_valid_eptp(u64 eptp)
--- a/arch/x86/kvm/vmx/pkvm/hyp/pgtable.c
+++ b/arch/x86/kvm/vmx/pkvm/hyp/pgtable.c
@@ -66,6 +66,19 @@ static bool leaf_mapping_allowed(struct pkvm_pgtable_ops *pgt_ops,
 	return leaf_mapping_valid(pgt_ops, vaddr, vaddr_end, pgsz_mask, level);
 }
 
+static void *pgtable_alloc_page(struct pkvm_mm_ops *mm_ops)
+{
+	void *page = NULL;
+
+	if (mm_ops->zalloc_page)
+		page = mm_ops->zalloc_page();
+
+	if (page && mm_ops->flush_cache)
+		mm_ops->flush_cache(page, PAGE_SIZE);
+
+	return page;
+}
+
 static void pgtable_set_entry(struct pkvm_pgtable_ops *pgt_ops,
 			struct pkvm_mm_ops *mm_ops,
 			void *ptep, u64 pte)
@@ -177,7 +190,7 @@ static int pgtable_map_walk_leaf(struct pkvm_pgtable *pgt,
 	 * page mapping. And for current level, if the huge page mapping is already
 	 * present, we need further split it.
 	 */
-	page = mm_ops->zalloc_page();
+	page = pgtable_alloc_page(mm_ops);
 	if (!page)
 		return -ENOMEM;
 
@@ -337,7 +350,7 @@ static int pgtable_unmap_cb(struct pkvm_pgtable *pgt, unsigned long vaddr,
 		 * if it is huge pte, split and goto next level.
 		 */
 		u64 prot = pgt_ops->pgt_entry_to_prot(ptep);
-		void *page = mm_ops->zalloc_page();
+		void *page = pgtable_alloc_page(mm_ops);
 
 		if (!page)
 			return -ENOMEM;
@@ -550,8 +563,8 @@ int pkvm_pgtable_init(struct pkvm_pgtable *pgt,
 	if (!mm_ops || !pgt_ops || !cap)
 		return -EINVAL;
 
-	if (alloc_root && mm_ops->zalloc_page) {
-		root = mm_ops->zalloc_page();
+	if (alloc_root) {
+		root = pgtable_alloc_page(mm_ops);
 		if (!root)
 			return -ENOMEM;
 		pgt->root_pa = __pkvm_pa(root);
--- a/arch/x86/kvm/vmx/pkvm/hyp/pkvm.c
+++ b/arch/x86/kvm/vmx/pkvm/hyp/pkvm.c
@@ -159,13 +159,8 @@ unsigned long __pkvm_teardown_shadow_vm(int shadow_vm_handle)
 
 	pkvm_pgstate_pgt_deinit(vm);
 
-	list_for_each_entry_safe(ptdev, tmp, &vm->ptdev_head, vm_node) {
-		pkvm_spin_lock(&vm->lock);
-		list_del(&ptdev->vm_node);
-		pkvm_spin_unlock(&vm->lock);
-
-		pkvm_detach_ptdev(ptdev);
-	}
+	list_for_each_entry_safe(ptdev, tmp, &vm->ptdev_head, vm_node)
+		pkvm_detach_ptdev(ptdev, vm);
 
 	shadow_size = vm->shadow_size;
 	memset(vm, 0, shadow_size);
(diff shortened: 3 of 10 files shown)


arch/x86/kvm/vmx/pkvm/hyp/pgtable.c
```diff
@@ -66,6 +66,19 @@ static bool leaf_mapping_allowed(struct pkvm_pgtable_ops *pgt_ops,
 	return leaf_mapping_valid(pgt_ops, vaddr, vaddr_end, pgsz_mask, level);
 }
 
+static void *pgtable_alloc_page(struct pkvm_mm_ops *mm_ops)
+{
+	void *page;
```
semihalf-maluka-dmytro, line 71: Uninitialized value of `page` may be used at line 76, as well as returned at line 79.
BTW I found this issue thanks to `-Werror,-Wsometimes-uninitialized` enabled in CrOS SDK build.

arch/x86/kvm/vmx/pkvm/hyp/ept.h
```diff
@@ -33,7 +33,7 @@ int pkvm_host_ept_init(struct pkvm_pgtable_cap *cap, void *ept_pool_base,
 		unsigned long ept_pool_pages);
 int handle_host_ept_violation(unsigned long gpa);
 void pkvm_flush_host_ept(void);
-int pkvm_shadow_ept_pool_init(void *ept_pool_base, unsigned long ept_pool_pages);
+int pkvm_shadow_pgt_pool_init(void *ept_pool_base, unsigned long ept_pool_pages);
```
semihalf-maluka-dmytro, line 36: Just a note: the name "shadow_pgt" is gonna be a bit ambiguous if we merge my patch https://chromium-review.googlesource.com/c/chromiumos/third_party/kernel/+/4231786 which adds a pool of shadow IOMMU page tables.

arch/x86/kvm/vmx/pkvm/hyp/pgtable.c
cxdong: Thanks. Will fix this.

arch/x86/kvm/vmx/pkvm/hyp/ept.h
cxdong: I see. The reason of changing to shadow_pgt is to make it common to both shadow ept and pgstate pgt as they are using the same pool which is shadow ept pool. With introducing iommu shadow page table pool by your path, shadow_pgt seems also common. In this case, do you think if keep the current name is better for this function?
BTW, I think keep using "shadow_pgt" inside ept.c should have no ambiguous. What do you think?

semihalf-maluka-dmytro: Yes, keeping using "shadow_pgt" for private functions inside ept.c seems perfectly fine. Using it for this public pkvm_shadow_pgt_pool_init() function would be slightly ambiguous with regard to shadow iommu page tables from my patch, but it is rather fine with me too (I have no better suggestions anyway), so please feel free to ignore  this remark.

I'm not sure what exactly do you mean by "With introducing iommu shadow page table pool by your path, shadow_pgt seems also common". Maybe you are suggesting to extend this shadow_pgt pool to use it for allocating shadow iommu page tables as well, together with shadow ept and pgstate pgt, instead of adding a new pool like in my patch? It seems not so important whether we use a separate pool or not, that said, unlike shadow_ept and pgstate_pgt, shadow iommu page tables are not tied with shadow VMs, since they are meant precisely for ptdevs *not* attached to any shadow VM.

arch/x86/kvm/vmx/pkvm/hyp/pkvm.c
```diff
@@ -157,15 +157,10 @@ unsigned long __pkvm_teardown_shadow_vm(int shadow_vm_handle)
...
-	list_for_each_entry_safe(ptdev, tmp, &vm->ptdev_head, vm_node) {
-		pkvm_spin_lock(&vm->lock);
-		list_del(&ptdev->vm_node);
-		pkvm_spin_unlock(&vm->lock);
+	list_for_each_entry_safe(ptdev, tmp, &vm->ptdev_head, vm_node)
+		pkvm_detach_ptdev(ptdev, vm);
 
-		pkvm_detach_ptdev(ptdev);
-	}
+	pkvm_pgstate_pgt_deinit(vm);
```
semihalf-maluka-dmytro, line 163: Again, just a note, not a suggestion: I understand you moved `pkvm_pgstate_pgt_deinit()` because `pkvm_pgstate_pgt_update_coherency()` should not be called for an already deinitialized pgstate_pgt, and it's probably ok per se, but just FYI it conflicts with my implementation of shadow iommu page tables for legacy mode as I explained in https://chromium-review.googlesource.com/c/chromiumos/third_party/kernel/+/4233618/4/arch/x86/kvm/vmx/pkvm/hyp/iommu.c#381. So I will need to rework it somehow.

cxdong: I see. So in legacy mode, detaching will sync ptdev->vpgt to ptdev->pgt, but because the mapping in vpgt is not undonated yet, the sync in legacy mode will be failed. But seems this failure is harmless because later host VMM will attach these devices to another IOMMU domain which will trigger IOMMU to sync again. So probably just need to make the ptdev->pgt sync stopped once there is undonated page?

arch/x86/kvm/vmx/pkvm/hyp/ept.h
cxdong: For the comment "With introducing iommu shadow page table pool by your path, shadow_pgt seems also common", as you mentioned the name of shadow_pgt is a bit ambiguous with your patch, I though you mean shadow_pgt is common to shadow_ept/pgstate_pgt/shadow_iommu_pgt, so pkvm_shadow_pgt_pool_init may make people think this function is used for shadow_iommu_pgt as well.

I can make the change to keep the name of pkvm_shadow_ept_pool_init(), so that there is no ambiguous.

JasonChenCJ: I think we shall try to unify these two pgtables, pgstate_pgt is designed as pgstate table and can be reused as IOMMU page table. For legacy IOMMU support, could we use it as host VM's pgstate table instead of host ept?

JasonChenCJ: After 2nd think, seemed for untrusted devs, it's not able to use IOMMU page table work as pgstate table - as there could be multiple untrusted devices with multiple IOMMU page tables. 
But still, we should be able to unify IOMMU shadow page table into shadow pgt pool, right?

semihalf-maluka-dmytro: @cxdong,
> I though you mean shadow_pgt is common to shadow_ept/pgstate_pgt/shadow_iommu_pgt, so pkvm_shadow_pgt_pool_init may make people think this function is used for shadow_iommu_pgt as well.

Right, that is what I meant. Agree, keeping the old name pkvm_shadow_ept_pool_init() would avoid this ambiguity (I don't insist on it though).

@JasonChenCJ,
Yes, there may be an arbitrary number of untrusted devices using different vIOMMU page tables, independently of the number of shadow VMs, so unifying it with pgstate_pgt doesn't sound quite feasible.

Regarding which pool to use for IOMMU shadow page table, I see such options:

1. add a new pool (as in my patches)
2. extend the shadow pgt pool (since these pgtables are a kind of shadow pgt)
3. extend the iommu pool (since these pgtables are a part iommu functionality)

and IIUC, it's just a matter of taste which of these 3 options to choose, it would not have an impact on any functionality. Please let me know if you think some of these options is better than others.

arch/x86/kvm/vmx/pkvm/hyp/pkvm.c
semihalf-maluka-dmytro: Yes, moving deinit before detaching (or maybe some other way to ensure the undonate is done before the iommu sync during detach) seems easier and cleaner to me.

arch/x86/kvm/vmx/pkvm/hyp/pgtable.c
cxdong: Fixed in the new version of pull request.

arch/x86/kvm/vmx/pkvm/hyp/pkvm.c
cxdong: Deinit has been moved before detaching in the new version of the pull request

arch/x86/kvm/vmx/pkvm/hyp/ept.h
JasonChenCJ: I prefer to #2.

semihalf-maluka-dmytro: @JasonChenCJ Ok, just to confirm: you'd prefer if I rework shadow iommu page tables in my patches, to use the shadow ept pool instead of adding a new pool, correct?
Now it seem like a good idea to me too: I can reuse the same mm_ops as for pgstate_pgt instead of reimplementing mm_ops for shadow iommu page tables, and also can reuse pkvm_pgstate_pgt_update_coherency().

JasonChenCJ: @semihalf-maluka-dmytro , yes, we can reuse what we had now. Thanks!

PR #21, opened 2023-03-03
Shadow IOMMU page tables for legacy IOMMU

semihalf-maluka-dmytro:
Unlike scalable IOMMU, legacy IOMMU provides just one level of address translation. So in the cases when the host itself uses address translation for the given device (i.e. uses legacy IOMMU in non-passthrough mode), pKVM needs to shadow the host's vIOMMU page tables to ensure memory protection. This PR implements shadow IOMMU page tables and enables using them in legacy mode when needed (i.e. when a device is used by the host in non-passthrough mode and is not attached to a protected VM), replacing our temporary hack using host's vIOMMU page tables directly.

This PR still does not fully ensure memory protection, since it doesn't implement syncing of shadow page tables immediately after a page is donated. Protection is ensured only after the host itself triggers IOMMU sync some time later.

v3:
- Fixed lack of `spgt->noncoherent_count` initialization.
- Removed `ptdev->spgt` and reworked `ptdev->pgt` handling accordingly.
- Added TODO about code refactoring for all the types of shadow 2nd level iommu page tables.
- Improved TODO about not reserving memory if there is no legacy IOMMU.
- Corrected commit message (added TODO) about lack of shadow iommu pgtable sync after page donate.

v2:
- Split the 1st patch: reserve memory for iommu_spgt in one patch, implement iommu_spgt API in the next one.
- Renamed `pgstate_pgt_mm_ops` to `shadow_sl_iommu_pgt_mm_ops` and cleaned up its usage correspondingly.
- Renamed `pkvm_shadow_iommu_pgtable_pages` to `pkvm_host_shadow_iommu_pgtable_pages`.
- Renamed `pkvm_{get,put}_iommu_spgt` to `pkvm_{get,put}_host_iommu_spgt`.
- Added TODO for not reserving memory if there is no legacy IOMMU.

v1:
Almost the same as the patchset submitted to ChromiumOS gerrit some time ago: https://chromium-review.googlesource.com/c/chromiumos/third_party/kernel/+/4233618 but with a couple of improvements:

- Instead of adding a separate memory pool for shadow IOMMU page tables, use (extend) the memory pool used for shadow EPT and pgstate tables (as suggested by @JasonChenCJ) which allows to reuse `pgstate_pgt_mm_ops` instead of implementing new mm_ops.
- Improved coherency handling: don't flush cache for the given shadow IOMMU page table if all the ptdevs using this page table are behind cache-coherent IOMMUs, even if some other IOMMUs are not cache-coherent (inspired by the implementation in PR #19 )

### DIFF
--- a/arch/x86/include/asm/kvm_pkvm.h
+++ b/arch/x86/include/asm/kvm_pkvm.h
@@ -200,6 +200,33 @@ static inline unsigned long pkvm_shadow_ept_pgtable_pages(int nr_vm)
 	return (res * 2);
 }
 
+/*
+ * Calculate the total pages for shadow IOMMU page tables for the host's
+ * devices used with Legacy IOMMU. Similarly to the calculation for shadow EPT,
+ * we assume that there is no shared memory between devices using different
+ * page tables.
+ *
+ * TODO: do not reserve these pages if legacy mode is not used by pKVM, i.e.
+ * if all the IOMMUs have scalable mode capability.
+ */
+static inline unsigned long pkvm_host_shadow_iommu_pgtable_pages(int nr_pdev)
+{
+	unsigned long pgtable_pages = __pkvm_pgtable_total_pages();
+	unsigned long res;
+
+	res = pgtable_pages;
+
+	/*
+	 * Similarly to shadow VMs (see the comment in
+	 * pkvm_shadow_ept_pgtable_pages()), each device may require
+	 * its own level2:level5 page table pages.
+	 */
+	res += __pkvm_pgtable_max_pages(pgtable_pages) * (nr_pdev - 1);
+
+	return res;
+}
+
+
 u64 pkvm_total_reserve_pages(void);
 
 int pkvm_init_shadow_vm(struct kvm *kvm);
--- a/arch/x86/kvm/vmx/pkvm/hyp/ept.c
+++ b/arch/x86/kvm/vmx/pkvm/hyp/ept.c
@@ -255,6 +255,12 @@ int pkvm_host_ept_unmap(unsigned long vaddr_start, unsigned long phys_start,
 	return pkvm_pgtable_unmap_safe(&host_ept, vaddr_start, phys_start, size, NULL);
 }
 
+void pkvm_host_ept_lookup(unsigned long vaddr, unsigned long *pphys,
+			  u64 *pprot, int *plevel)
+{
+	pkvm_pgtable_lookup(&host_ept, vaddr, pphys, pprot, plevel);
+}
+
 void pkvm_host_ept_destroy(void)
 {
 	pkvm_pgtable_destroy(&host_ept, NULL);
--- a/arch/x86/kvm/vmx/pkvm/hyp/ept.c
+++ b/arch/x86/kvm/vmx/pkvm/hyp/ept.c
@@ -482,15 +482,32 @@ static struct pkvm_mm_ops shadow_ept_mm_ops = {
 };
 
 /*
- * pgstate_pgt_mm_ops is similar to the shadow_ept_mm_ops as its
- * memory is reserved together with shadow EPT pages. The
- * difference is that it doesn't have the flush_tlb callback as
- * pgstate_pgt only works as IOMMU second-level page table for
- * protected VM when there are passthrough devices, and in this
- * case the memory is pinned, and the mapping is not allowed to
- * be removed from pgstate_pgt.
+ * mm_ops for shadow second-level IOMMU page tables. These tables
+ * are similar to shadow EPT tables, as they also have the EPT
+ * format and their memory is reserved together with shadow EPT
+ * pages. The difference is that this mm_ops doesn't have the
+ * flush_tlb callback.
+ *
+ * Precisely, shadow_sl_iommu_pgt_mm_ops is used for two kinds of
+ * 2nd level iommu page tables:
+ *
+ * - pgstate_pgt which is reused as IOMMU page table for protected
+ *   VM with passthrough devices. In this case the memory is pinned,
+ *   and the mapping is not allowed to be removed from pgstate_pgt,
+ *   so the flush_tlb callback is not needed.
+ *
+ * - Host shadow IOMMU page tables used for the host's devices when
+ *   legacy IOMMU is used. They do not need the flush_tlb callback
+ *   either, since IOTLB flush after unmapping pages from these
+ *   tables is performed in other ways: either as a part of vIOMMU
+ *   IOTLB flush emulation when initiated by the host, or together
+ *   with host EPT TLB flush when ensuring pKVM memory protection.
+ *
+ * TODO: refactor the code: move all the management of both types
+ * of 2nd level iommu page tables to iommu_spgt.c to some common API.
+ * That means also refactoring of pkvm_ptdev structure.
  */
-static struct pkvm_mm_ops pgstate_pgt_mm_ops = {
+static struct pkvm_mm_ops shadow_sl_iommu_pgt_mm_ops = {
 	.phys_to_virt = pkvm_phys_to_virt,
 	.virt_to_phys = pkvm_virt_to_phys,
 	.zalloc_page = shadow_pgt_zalloc_page,
@@ -501,12 +518,12 @@ static struct pkvm_mm_ops pgstate_pgt_mm_ops = {
 };
 
 /*
- * As pgstate pgt may be used as IOMMU page table, flushing cache is
- * needed when modifying the page table entries if IOMMU is not coherent.
- * This ops has flush_cache callback and can be used for the pgstate_pgt
- * which is used as IOMMU page table with noncoherent IOMMU.
+ * Flushing cache is needed when modifying IOMMU page table entries
+ * if the IOMMU is not coherent. This ops has flush_cache callback
+ * so it can be used for a pgtable which is used as IOMMU page table
+ * with noncoherent IOMMU.
  */
-static struct pkvm_mm_ops pgstate_pgt_mm_ops_noncoherency = {
+static struct pkvm_mm_ops shadow_sl_iommu_pgt_mm_ops_noncoherency = {
 	.phys_to_virt = pkvm_phys_to_virt,
 	.virt_to_phys = pkvm_virt_to_phys,
 	.zalloc_page = shadow_pgt_zalloc_page,
@@ -753,15 +770,21 @@ int pkvm_pgstate_pgt_init(struct pkvm_shadow_vm *vm)
 		.table_prot = VMX_EPT_RWX_MASK,
 	};
 
-	return pkvm_pgtable_init(pgt, &pgstate_pgt_mm_ops, &ept_ops, &cap, true);
+	return pkvm_pgtable_init(pgt, &shadow_sl_iommu_pgt_mm_ops, &ept_ops, &cap, true);
 }
 
-void pkvm_pgstate_pgt_update_coherency(struct pkvm_pgtable *pgt, bool coherent)
+struct pkvm_mm_ops *pkvm_shadow_sl_iommu_pgt_get_mm_ops(bool coherent)
+{
+	return coherent ? &shadow_sl_iommu_pgt_mm_ops
+			: &shadow_sl_iommu_pgt_mm_ops_noncoherency;
+}
+
+void pkvm_shadow_sl_iommu_pgt_update_coherency(struct pkvm_pgtable *pgt, bool coherent)
 {
 	if (coherent)
-		pkvm_pgtable_set_mm_ops(pgt, &pgstate_pgt_mm_ops);
+		pkvm_pgtable_set_mm_ops(pgt, &shadow_sl_iommu_pgt_mm_ops);
 	else
-		pkvm_pgtable_set_mm_ops(pgt, &pgstate_pgt_mm_ops_noncoherency);
+		pkvm_pgtable_set_mm_ops(pgt, &shadow_sl_iommu_pgt_mm_ops_noncoherency);
 }
 
 /*
--- a/arch/x86/kvm/vmx/pkvm/hyp/iommu.c
+++ b/arch/x86/kvm/vmx/pkvm/hyp/iommu.c
@@ -414,7 +414,7 @@ static bool sync_shadow_context_entry(struct pgt_sync_data *sdata)
 
 		if (!sdata->guest_ptep) {
 			if (context_lm_is_present(shadow_ce)) {
-				pkvm_setup_ptdev_vpgt(ptdev, 0, NULL, NULL, NULL);
+				pkvm_setup_ptdev_vpgt(ptdev, 0, NULL, NULL, NULL, false);
 				pkvm_setup_ptdev_did(ptdev, 0);
 				iommu_del_ptdev(iommu, ptdev);
 
@@ -431,7 +431,7 @@ static bool sync_shadow_context_entry(struct pgt_sync_data *sdata)
 			if (aw != 1 && aw != 2 && aw != 3) {
 				pkvm_err("pkvm: unsupported address width %u\n", aw);
 
-				pkvm_setup_ptdev_vpgt(ptdev, 0, NULL, NULL, NULL);
+				pkvm_setup_ptdev_vpgt(ptdev, 0, NULL, NULL, NULL, false);
 				pkvm_setup_ptdev_did(ptdev, 0);
 
 				/*
@@ -446,7 +446,7 @@ static bool sync_shadow_context_entry(struct pgt_sync_data *sdata)
 				    (aw == 2) ? 4 : 5;
 			cap.allowed_pgsz = pkvm_hyp->ept_cap.allowed_pgsz;
 			pkvm_setup_ptdev_vpgt(ptdev, context_lm_get_slptr(guest_ce),
-					      &viommu_mm_ops, &ept_ops, &cap);
+					      &viommu_mm_ops, &ept_ops, &cap, false);
 
 			break;
 		case CONTEXT_TT_PASS_THROUGH:
@@ -460,7 +460,7 @@ static bool sync_shadow_context_entry(struct pgt_sync_data *sdata)
 		default:
 			pkvm_err("pkvm: unsupported translation type %u\n", tt);
 
-			pkvm_setup_ptdev_vpgt(ptdev, 0, NULL, NULL, NULL);
+			pkvm_setup_ptdev_vpgt(ptdev, 0, NULL, NULL, NULL, false);
 			pkvm_setup_ptdev_did(ptdev, 0);
 			goto update_shadow_ce;
 		}
@@ -559,7 +559,7 @@ static bool sync_shadow_pasid_table_entry(struct pgt_sync_data *sdata)
 			 * a ptdev's vpgt/did should be reset as well as
 			 * deleting ptdev from this iommu.
 			 */
-			pkvm_setup_ptdev_vpgt(ptdev, 0, NULL, NULL, NULL);
+			pkvm_setup_ptdev_vpgt(ptdev, 0, NULL, NULL, NULL, false);
 			pkvm_setup_ptdev_did(ptdev, 0);
 			iommu_del_ptdev(iommu, ptdev);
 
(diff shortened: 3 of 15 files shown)


arch/x86/include/asm/kvm_pkvm.h
```diff
@@ -200,6 +200,29 @@ static inline unsigned long pkvm_shadow_ept_pgtable_pages(int nr_vm)
...
+	unsigned long res;
+
+	res = pgtable_pages;
+
+	/*
+	 * Similarly to shadow VMs (see the comment in
+	 * pkvm_shadow_ept_pgtable_pages()), each device may require
+	 * its own level2:level5 page table pages.
+	 */
+	res += __pkvm_pgtable_max_pages(pgtable_pages) * (nr_pdev - 1);
```
semihalf-maluka-dmytro, line 220: Is this enough or maybe we need twice as many pages to support vSMM mode like in `pkvm_shadow_ept_pgtable_pages()`? (I know nothing about vSMM.)

arch/x86/kvm/vmx/pkvm/hyp/ept.c
```diff
@@ -489,8 +489,16 @@ static struct pkvm_mm_ops shadow_ept_mm_ops = {
  * protected VM when there are passthrough devices, and in this
  * case the memory is pinned, and the mapping is not allowed to
  * be removed from pgstate_pgt.
+ *
+ * Also pgstate_pgt_mm_ops is reused for shadow IOMMU page tables
```
semihalf-maluka-dmytro, line 499: As `pgstate_pgt_mm_ops` becomes used not just for pgstate_pgt but also for iommu shadow pgt, maybe we should rename it to avoid confusion? But I'm not sure what name to use to describe both pgstate_pgt and iommu_spgt and at the same time distinguish them both from shadow_ept. Perhaps just `shadow_ept_mm_ops_notlbflush` (and correspondingly `shadow_ept_mm_ops_notlbflush_noncoherency`)?

arch/x86/kvm/vmx/pkvm/hyp/ptdev.c
```diff
@@ -71,22 +73,41 @@ void pkvm_put_ptdev(struct pkvm_ptdev *ptdev)
...
+
 	memset(ptdev, 0, sizeof(struct pkvm_ptdev));
 
 	pkvm_spin_unlock(&ptdev_lock);
 }
 
 void pkvm_setup_ptdev_vpgt(struct pkvm_ptdev *ptdev, unsigned long root_gpa,
 			   struct pkvm_mm_ops *mm_ops, struct pkvm_pgtable_ops *paging_ops,
-			   struct pkvm_pgtable_cap *cap)
+			   struct pkvm_pgtable_cap *cap, bool shadowed)
```
JasonChenCJ, line 86: Shadow iommu pgt setup should not be in pkvm_setup_ptdev_vpgt. It's kind of setup ptdev_pgt.

To me, a ptdev may be under different situations:
1. ptdev belong to host
    - scalable mode: ptdev->vpgt = host_set_flptr, ptdev->pgt = host_ept
    - legacy mode (type=PT): ptdev->vpgt = NULL, ptdev->pgt = host_ept
    - legacy mode (type!=PT): ptdev->vpgt = host_set_slptr, ptdev->pgt = get_iommu_spgt
2. ptdev belong to a guest
   - scalable mode: ptdev->vpgt = host_set_flptr, ptdev->pgt = pgstate_pgt
   - legacy mode: ptdev->vpgt = host_set_slptr, ptdev->pgt = pgstate_pgt

I am not sure if we can do according to above, no new added ptdev->spgt, but just reuse ptdev->pgt.

arch/x86/include/asm/kvm_pkvm.h
semihalf-maluka-dmytro: > I don't think vSMM will use pass-thru devices.

Cool.

> BTW, here the shadow iommu pgtable is actually for host VM, could it be pkvm_host_shadow_iommu_pgtable_pages?
(I have to say pgstate table is also a shadow iommu pgtable, but it's per L2 VM. Here shadow iommu pgtable is specific for host VM.)

To be precise, shadow iommu pgtable may be also used for normal VM. But normal VM devices are still "owned" by the host VM (not attached to any L2 VM), so I'm ok with renaming this to `pkvm_host_shadow_iommu_pgtable_pages`.

> And could we make it be 0 if IOMMU is scalable mode?

That would be nice (to save memory) but the problem is that, as I understand, this calculation is used by the host in `pkvm_reserve()` which is done very early, before we know anything about IOMMUs.
We can add a Kconfig option for disable legacy IOMMU support, OTOH it probably would be good to avoid adding compilation options for things that can be detected automatically.
So I was assuming that we can reserve this memory unconditionally for now, and in the future, if needed, think how to disable reserving it if there is no legacy IOMMU.

arch/x86/kvm/vmx/pkvm/hyp/ept.c
semihalf-maluka-dmytro: Ok, I think I like this idea. It reflects the fact that both tables have EPT format.

arch/x86/kvm/vmx/pkvm/hyp/iommu_spgt.c
```diff
@@ -0,0 +1,109 @@
...
+#include "ept.h"
+#include "bug.h"
+
+static DEFINE_HASHTABLE(iommu_spgt_hasht, 8);
+static DECLARE_BITMAP(iommu_spgt_bitmap, PKVM_MAX_PDEV_NUM);
+static struct pkvm_iommu_spgt pkvm_iommu_spgt[PKVM_MAX_PDEV_NUM];
+static pkvm_spinlock_t iommu_spgt_lock = { __ARCH_PKVM_SPINLOCK_UNLOCKED };
+
+struct pkvm_pgtable *pkvm_get_iommu_spgt(unsigned long root_gpa, bool coherency)
+{
```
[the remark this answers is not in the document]
semihalf-maluka-dmytro: > pkvm_host_get_iommu_spgt() as it's only for host VM?

I'm fine with it, OTOH maybe this get/put API does not need to be aware of the fact that it's used only for the host?

> BTW, this patch could be separated into two patches, one for memory reservation, another one for adding host shadow iommu page table APIs.

Agree.

arch/x86/kvm/vmx/pkvm/hyp/ptdev.c
semihalf-maluka-dmytro: > Shadow iommu pgt setup should not be in pkvm_setup_ptdev_vpgt. It's kind of setup ptdev_pgt.

Agree, seems like a good idea.

Regarding ptdev->spgt, it is used to remember the currently used shadow iommu pgtable when ptdev is attached to a VM and its pgt is thus changed to pgstate_pgt. When the ptdev is detached, its pgt is changed back to the remembered spgt.
Probably we can get rid of ptdev->spgt if we just do pkvm_put_iommu_spgt (i.e. free the shadow iommu pgtable) when attaching, and then pkvm_get_iommu_spgt (i.e. allocate a new shadow iommu pgtable) when detaching. We would also need to check during detaching if we need a shadow pgtable or not, and to sync the newly allocated shadow pgtable with vpgt (OTOH perhaps that would be automatically ensured by pkvm_iommu_sync() during detach). Do you think it would be a better solution than the current one with ptdev->spgt?

arch/x86/include/asm/kvm_pkvm.h
JasonChenCJ: > > BTW, here the shadow iommu pgtable is actually for host VM, could it be pkvm_host_shadow_iommu_pgtable_pages?
> > (I have to say pgstate table is also a shadow iommu pgtable, but it's per L2 VM. Here shadow iommu pgtable is specific for host VM.)
> 
> To be precise, shadow iommu pgtable may be also used for normal VM. But normal VM devices are still "owned" by the host VM (not attached to any L2 VM), so I'm ok with renaming this to `pkvm_host_shadow_iommu_pgtable_pages`.
> 

Yes, normal VM's devices are still owned by the host. But in our design, for normal VM, we are using pgstate table as the shadow iommu pgtable. It's different with this one.

> > And could we make it be 0 if IOMMU is scalable mode?
> 
> That would be nice (to save memory) but the problem is that, as I understand, this calculation is used by the host in `pkvm_reserve()` which is done very early, before we know anything about IOMMUs. We can add a Kconfig option for disable legacy IOMMU support, OTOH it probably would be good to avoid adding compilation options for things that can be detected automatically. So I was assuming that we can reserve this memory unconditionally for now, and in the future, if needed, think how to disable reserving it if there is no legacy IOMMU.

Got it, I am okay for it.

arch/x86/kvm/vmx/pkvm/hyp/iommu_spgt.c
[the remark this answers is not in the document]
JasonChenCJ: > > pkvm_host_get_iommu_spgt() as it's only for host VM?
> 
> I'm fine with it, OTOH maybe this get/put API does not need to be aware of the fact that it's used only for the host?
> 

Yes, I agree. If so, could we combine it with pgstate_pgt? The pgstate pgt should also be able to allocate from this API.

arch/x86/kvm/vmx/pkvm/hyp/ptdev.c
JasonChenCJ: > > Shadow iommu pgt setup should not be in pkvm_setup_ptdev_vpgt. It's kind of setup ptdev_pgt.
> 
> Agree, seems like a good idea.
> 
> Regarding ptdev->spgt, it is used to remember the currently used shadow iommu pgtable when ptdev is attached to a VM and its pgt is thus changed to pgstate_pgt. When the ptdev is detached, its pgt is changed back to the remembered spgt. Probably we can get rid of ptdev->spgt if we just do pkvm_put_iommu_spgt (i.e. free the shadow iommu pgtable) when attaching, and then pkvm_get_iommu_spgt (i.e. allocate a new shadow iommu pgtable) when detaching. We would also need to check during detaching if we need a shadow pgtable or not, and to sync the newly allocated shadow pgtable with vpgt (OTOH perhaps that would be automatically ensured by pkvm_iommu_sync() during detach). Do you think it would be a better solution than the current one with ptdev->spgt?

I suppose host VM will not keep old ptdev->vpgt when this ptdev be unbind then assigned to a VM. So I don't think we should record old ptdev->spgt. We can check if the unbind will also clear the slptr in the context table entry, if so, you even do not need call put_iommU_spgt during attaching.

arch/x86/include/asm/kvm_pkvm.h
semihalf-maluka-dmytro: > But in our design, for normal VM, we are using pgstate table as the shadow iommu pgtable. It's different with this one.

Hmm, don't we use pgstate table as the iommu pgtable for protected VMs only? ptdev->pgt is set to pgstate_pgt only in pkvm_attach_ptdev(), which is called only for protected VM. Maybe I'm misunderstanding your point.

arch/x86/kvm/vmx/pkvm/hyp/iommu_spgt.c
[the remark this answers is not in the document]
semihalf-maluka-dmytro: Hmm, interesting idea. But it seems we cannot just adapt this API for pgstate_pgt as is, since it assumes 1:1 correspondence between vIOMMU page table (identified by its root address) and shadow IOMMU page table, while pgstate_pgt is per VM, and as I understand, each VM needs its own pgstate_pgt for storing its own page states, even if different VMs share the same vIOMMU page table.

So now I'm thinking, maybe it's indeed better to name it pkvm_host_get_iommu_spgt() as you suggest, since this API as is seems suitable for the host only.

arch/x86/include/asm/kvm_pkvm.h
JasonChenCJ: > > But in our design, for normal VM, we are using pgstate table as the shadow iommu pgtable. It's different with this one.
> 
> Hmm, don't we use pgstate table as the iommu pgtable for protected VMs only? ptdev->pgt is set to pgstate_pgt only in pkvm_attach_ptdev(), which is called only for protected VM. Maybe I'm misunderstanding your point.

For normal VM, we always use pgstate table to maintain page state.
And for normal VM under scalable IOMMU, we use host ept as IOMMU pgtable.
While for normal VM under legacy IOMMU, yes, we can use IOMMU spgt as IOMMU pgtable, but I think this spgt should combine with pgstate pgtable. Otherwise, we will have both IOMMU spgt and pgstate pgt.

arch/x86/kvm/vmx/pkvm/hyp/iommu_spgt.c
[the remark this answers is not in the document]
JasonChenCJ: After you know a ptdev belong to a guest VM, and know ptdev->pgt is pgstate pgt, then you can directly return pgstate pgt for this API - which means you will not touch your spgt logic.

Is it ok?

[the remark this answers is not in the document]
semihalf-maluka-dmytro: Are you suggesting to rework the way how we set ptdev->pgt to pgstate_pgt for an attached ptdev: instead of doing it in pkvm_attach_ptdev(), do it through this generalized get_iommu_spgt API which will decide whether the given ptdev shall use pgstate_pgt (if attached) or a specially allocated iommu spgt (if not attached)? If so, get_iommu_spgt would need access to the pkvm_shadow_vm structure (to get &vm->pgstate_pgt) which is currently not easily available through pkvm_ptdev structure.

Overall, this sounds like a significant reorganization of the existing code. And maybe in the end it would make the code cleaner, or maybe quite the opposite, it would make it just more confusing.

I was actually intending just to add support for legacy IOMMU shadow tables, without making drastic changes in the existing code for pgstate_pgt.

arch/x86/kvm/vmx/pkvm/hyp/ptdev.c
semihalf-maluka-dmytro: Regarding "Shadow iommu pgt setup should not be in pkvm_setup_ptdev_vpgt", actually I've recalled why I added it in pkvm_setup_ptdev_vpgt: we need to check if the vpgt changes or not (`root_gpa != ptdev->vpgt.root_pa`), since e.g. if the old vpgt was shared with another device and the new vpgt is not, we cannot just use the same shadow iommu pgt, we need to put the old shadow iommu pgt and get a new one for the new root_gpa. It seems we still need this check even if we get rid of ptdev->spgt. So it seems it's better to do this setup in pkvm_setup_ptdev_vpgt, right?

arch/x86/include/asm/kvm_pkvm.h
JasonChenCJ: I think currently we may keep both IOMMU spgt & pgstate pgt for normal VM under legacy IOMMU mode, it will not make your patch complicated as now you don't have page state control for IOMMU spgt at all. We can consider how to add page state control for all IOMMU spgt (include normal VM one) in the future.

arch/x86/kvm/vmx/pkvm/hyp/iommu_spgt.c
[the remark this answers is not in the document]
JasonChenCJ: Yes, I agree, currently we do not need to do big refactor for the code.

arch/x86/kvm/vmx/pkvm/hyp/ptdev.c
JasonChenCJ: Yes, after 2nd think, ptdev->pgt is actually a hide parameter in ptdev, I am ok to put its update in pkvm_setup_ptdev_vpgt.

arch/x86/include/asm/kvm_pkvm.h
semihalf-maluka-dmytro: - Renamed to `pkvm_host_shadow_iommu_pgtable_pages`
- Added TODO for not reserving memory if there is no legacy IOMMU

arch/x86/kvm/vmx/pkvm/hyp/ept.c
semihalf-maluka-dmytro: Renamed to `shadow_sl_iommu_pgt_mm_ops`.

arch/x86/kvm/vmx/pkvm/hyp/iommu_spgt.c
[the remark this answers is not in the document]
semihalf-maluka-dmytro: Renamed to `pkvm_get_host_iommu_spgt()` and `pkvm_put_host_iommu_spgt()`.

arch/x86/kvm/vmx/pkvm/hyp/ptdev.c
semihalf-maluka-dmytro: In the updated patches I decided to keep ptdev->spgt.
Actually I've also tried removing ptdev->spgt, in this PoC patch: https://github.com/semihalf-maluka-dmytro/pKVM-IA/commit/459f15245a32f0357a1d83863f3a81339a023679. It seems it work, but it seems it just makes the code more confusing.

semihalf-maluka-dmytro: After more thought, I think I'll remove ptdev->spgt right away (i.e. basically, apply https://github.com/semihalf-maluka-dmytro/pKVM-IA/commit/459f15245a32f0357a1d83863f3a81339a023679 to this PR): you are right, the host will typically not keep the old vpgt when attaching device to a VM, it will use a new vpgt, and there is no point in allocating additional ptdev->spgt table which will not be used at all (since pgstate_pgt will be used instead). That's independently of the bigger refactoring you've mentioned.

Regarding the bigger refactoring (moving all the management of both type of 2nd level iommu pgt to iommu_spgt.c to some common API and reworking ptdev): ok, I will add a TODO about it.

arch/x86/kvm/vmx/pkvm/hyp/ept.c
```diff
@@ -255,6 +255,12 @@ int pkvm_host_ept_unmap(unsigned long vaddr_start, unsigned long phys_start,
 	return pkvm_pgtable_unmap_safe(&host_ept, vaddr_start, phys_start, size, NULL);
 }
 
+void pkvm_host_ept_lookup(unsigned long vaddr, unsigned long *pphys,
```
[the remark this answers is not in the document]
semihalf-maluka-dmytro: Agree. Do you mean also adding TODO in the code, not only in the commit message? If so, where would you prefer to add such TODO? Perhaps in `host_initiate_donation()` in mem_protect.c?

arch/x86/kvm/vmx/pkvm/hyp/iommu.c
```diff
@@ -26,38 +26,38 @@ static struct pkvm_iommu iommus[PKVM_MAX_IOMMU_NUM];
...
 
 /*
- * Guest page table walking parameter.
+ * Guest root/context/pasid table (hereinafter "id table") walking parameter.
  * pkvm IOMMU driver walks the guest page table when syncing
- * with the shadow page table.
+ * with the shadow id table.
  */
-struct pgt_sync_walk_data {
+struct id_sync_walk_data {
```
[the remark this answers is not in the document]
semihalf-maluka-dmytro: Wouldn't it be confusing with Interrupt Descriptor Table?

arch/x86/kvm/vmx/pkvm/hyp/ptdev.c
semihalf-maluka-dmytro: Removed ptdev->spgt and added TODO about refactoring (in the comment for `shadow_sl_iommu_pgt_mm_ops` in ept.c).

arch/x86/kvm/vmx/pkvm/hyp/ept.c
```diff
@@ -482,15 +482,28 @@ static struct pkvm_mm_ops shadow_ept_mm_ops = {
...
+ *   VM with passthrough devices. In this case the memory is pinned,
+ *   and the mapping is not allowed to be removed from pgstate_pgt,
+ *   so the flush_tlb callback is not needed.
+ *
+ * - Host shadow IOMMU page tables used for the host's devices when
+ *   legacy IOMMU is used. They do not need the flush_tlb callback
+ *   either, since IOTLB flush after unmapping pages from these
+ *   tables is performed in other ways: either as a part of vIOMMU
+ *   IOTLB flush emulation when initiated by the host, or together
+ *   with host EPT TLB flush when ensuring pKVM memory protection.
```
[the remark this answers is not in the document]
semihalf-maluka-dmytro: Added TODO about future refactoring.

arch/x86/kvm/vmx/pkvm/hyp/ept.c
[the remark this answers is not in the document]
semihalf-maluka-dmytro: Added TODO in the commit message.

arch/x86/kvm/vmx/pkvm/hyp/iommu.c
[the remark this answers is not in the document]
JasonChenCJ: I don't think IOMMU will touch interrupt descriptor table, so in iommu.c, idt should not cause confuse? But if you insist, I am ok with id.

PR #22, opened 2023-04-21
Cfg space audit

tzhang449:
Add general PIO and MMIO emulation logic for host vm to emulate both
legacy and ECAM PCI config space access mechanism in pkvm. The config
space access for passthrough device is then intercepted and emulated
in pKVM. For now, host is not allowed to access the base address
registers if it belongs to a passthrough device. Add two new files,
io_emulate.c and pci.c to place I/O emulation and PCI related code.

Changelog:
----------
v2:
- Refactor and refine patch structure.
- Some vcpu fails to enter VMX non-root mode if they share IO bitmap, so
  allocate IO bitmap per vcpu.
- Compiler sometimes would use x86_64 new rigisters, which adds a REX
  prefix to the instruction opcode. Handle REX prefix in the MMIO simple
  decoder.
- ACPI driver also touches mmcfg region. Support movzbl in MMIO simple
  decoder to handle this.
- Use bitmapped array to store PIO handlers instead of linked list.
- Add pkvm_pci_info in host vm, so can store pci mmcfg list from host.
- Unmap all mmcfg regions from host EPT, so all mmcfg range is emulated.

v3:
- Change most IO emulation code to host only.
- Separate default PIO and MMIO handler to two new patches.
- Introduce "io.h" header to put IO related code.
- Change MMIO emulation handler table from list to bitmap.
- Refactor PCI cfg space access function to make it more clear.

v4:
- Cache config space address reg and only write the reg when
  data port is accessed.
- Narrow to deny BAR access for passthrough device.


arch/x86/kvm/vmx/pkvm/include/pkvm.h
```diff
@@ -111,7 +113,8 @@ struct pkvm_section {
 #define PKVM_PAGES (ALIGN(sizeof(struct pkvm_hyp), PAGE_SIZE) >> PAGE_SHIFT)
 #define PKVM_PCPU_PAGES (ALIGN(sizeof(struct pkvm_pcpu), PAGE_SIZE) >> PAGE_SHIFT)
 #define PKVM_HOST_VCPU_PAGES (ALIGN(sizeof(struct pkvm_host_vcpu), PAGE_SIZE) >> PAGE_SHIFT)
-#define PKVM_VMCS_PAGES 3 /*vmxarea+vmcs+msr_bitmap*/
+#define PKVM_HOST_VCPU_VMCS_PAGES 3 /*vmxarea+vmcs+msr_bitmap*/
+#define PKVM_EXTRA_PAGES 2 /*host vm io bitmap*/
```
JasonChenCJ: can it combine with PKVM_PAGES?

arch/x86/kvm/vmx/pkvm/hyp/io_emulate.h
```diff
@@ -32,4 +32,7 @@ struct pkvm_pio_emul_table {
 	DECLARE_BITMAP(bitmap, PKVM_MAX_PIO_EMUL_NUM);
 };
 
+bool pio_access_valid(int size);
```
JasonChenCJ, line 61: do we need expose pio_access_valid?

arch/x86/kvm/vmx/pkvm/hyp/io_emulate.h
```diff
@@ -32,6 +32,8 @@ struct pkvm_pio_emul_table {
 	DECLARE_BITMAP(bitmap, PKVM_MAX_PIO_EMUL_NUM);
 };
 
+int register_host_pio_handler(struct pkvm_host_vm *host_vm, unsigned int port,
+	unsigned int size_mask, pio_handler_t read, pio_handler_t write);
```
JasonChenCJ, line 60: we may expose this at first patch it come out.

arch/x86/kvm/vmx/pkvm/include/pkvm.h
tzhang449: PKVM_PAGES is used to allocate pages for pkvm_hyp structure. Beacuse the io bitmap and the later mmcfg list table is not added to pkvm_hyp, don't think we can use it?

arch/x86/kvm/vmx/pkvm/hyp/io_emulate.h
tzhang449: Sorry this is not needed, will change to static.

arch/x86/kvm/vmx/pkvm/include/pkvm.h
JasonChenCJ: Okay, got it.

arch/x86/kvm/vmx/pkvm/hyp/io_emulate.h
tzhang449: I actually did this. Static in [pkvm: introduce port I/O emulation framework](https://github.com/intel-staging/pKVM-IA/pull/22/commits/8f6f66622a61328621267177fb73fde699c3f6f9) and exposed in [pkvm: emulate PCI config space access](https://github.com/intel-staging/pKVM-IA/pull/22/commits/e232085fc6152e8499804d4ed7108f704fdac219).

JasonChenCJ: what I mean is you can expose it in [pkvm: introduce port I/O emulation framework](https://github.com/intel-staging/pKVM-IA/pull/22/commits/8f6f66622a61328621267177fb73fde699c3f6f9), but I am ok if you keep current solution.

tzhang449: Got it, I will change.

arch/x86/kvm/vmx/pkvm/hyp/io_emulate.h
tzhang449: Changed to static

arch/x86/kvm/vmx/pkvm/hyp/io_emulate.h
tzhang449: Fixed it.

arch/x86/kvm/vmx/pkvm/hyp/io_emulate.c
```diff
@@ -108,7 +108,7 @@ static int emulate_host_pio(struct kvm_vcpu *vcpu, struct pkvm_pio_req *req)
 	return ret;
 }
 
-bool pio_access_valid(int size)
+static bool pio_access_valid(int size)
```
semihalf-maluka-dmytro, line 111: `error: static declaration of 'pio_access_valid' follows non-static declaration`

So it's better to move `pio_access_valid()` before `get_pio_handler()`.

Besides that, look like `pio_access_valid()` should be made static already in https://github.com/intel-staging/pKVM-IA/pull/22/commits/0a0b5bb10c183308988ab0303c8821b262a7f192 where it is introduced,  not in this patch which is about MMIO.

arch/x86/kvm/vmx/pkvm/hyp/pci.c
```diff
@@ -52,6 +52,18 @@ static int pci_mmcfg_write(u64 address, int size, unsigned long value)
 	return 0;
 }
 
+static bool host_vpci_cfg_space_audit_access(struct pkvm_ptdev *ptdev, u64 offset)
+{
+	if (!ptdev || ptdev->shadow_vm_handle == 0)
```
semihalf-maluka-dmytro, line 57: Don't we race with ptdev attach/detach here?

arch/x86/kvm/vmx/pkvm/hyp/pci.c
```diff
@@ -52,6 +52,18 @@ static int pci_mmcfg_write(u64 address, int size, unsigned long value)
...
+		return true;
+
+	/* Deny for all BAR access */
+	if (offset >= 0x10 && offset < 0x28)
+		return false;
```
semihalf-maluka-dmytro, line 62: I suppose that in the final solution we will need audit only for BAR write, not for BAR read? Maybe let's not deny BAR read access?

BTW, with this PR, I ran a Linux VM via crosvm as a protected VM with iwlwifi passed-thru, and iwlwifi was failing in the VM:
```
[    1.035269] iwlwifi 0000:00:14.0: HW_REV=0xFFFFFFFF, PCI issues?
[    1.035671] iwlwifi: probe of 0000:00:14.0 failed with error -5
```

Even after I commented out audit for BAR read, it still wasn't working. Only after I commented out audit for BAR write as well, iwlwifi started working again.

So at least in some setups this PR makes a protected VM's passthrough device non-workable (i.e. the VMM needs access to physical BARs even after attaching the device to the VM).

JasonChenCJ: For vfio-pci pass-thru, the BAR should be totally emulated, it means guest BAR program should not finally touch real ones. We need ensure what's the behavior host will do for BAR programing after assign a device.

semihalf-maluka-dmytro: Yep, I know, I meant BAR access by the host (which, though, apparently affects the guest's emulated PCI cfg space functionality too). I was observing a few dozen `pkvm: handle host port I/O access failed` errors at the moment of the VM startup. I can debug why exactly crosvm does that.

JasonChenCJ: During VM startup, BAR reading from host is possible, but what's the reason of BAR writing?

arch/x86/kvm/vmx/pkvm/hyp/io_emulate.c
tzhang449: Sorry my fault. Fixed it.

arch/x86/kvm/vmx/pkvm/hyp/pci.c
tzhang449: Yeah this is a race condition. I updated this function to acquire ptdev->lock.

arch/x86/kvm/vmx/pkvm/hyp/pci.c
tzhang449: I updated the patchset to allow BAR read.

For the BAR writing behavior dmytro you notice, I believe this is done when crosvm create the vfio device:
crosvm call VFIO_GROUP_GET_DEVICE_FD ->
vfio_group_fops_unl_ioctl->
vfio_group_get_device_fd->
vfio_device_open->
vfio_pci_open_device->
vfio_pci_core_enable->
pci_try_reset_function->
pci_dev_restore

So the crosvm first call KVM_CREATE_DEVICE, where pKVM add the device to shadow vm.
Then crosvm calls VFIO_GROUP_GET_DEVICE_FD, which would reset the pci function and trigger BAR write.

I make a test to reverse the two in crosvm, ie. first get the vfio fd then create the kvm device, it works. No BAR write is seen after the kvm device is created.
Maybe dmytro you could have a try with this?

JasonChenCJ: @tzhang449 , do you mean BAR reading instead of BAR writing here for VFIO_GROUP_GET_DEVICE_FD?

> So the crosvm first call KVM_CREATE_DEVICE, where pKVM add the device to shadow vm. Then crosvm calls VFIO_GROUP_GET_DEVICE_FD, which would reset the pci function and trigger BAR write.
> 
> I make a test to reverse the two in crosvm, ie. first get the vfio fd then create the kvm device, it works. No BAR write is seen after the kvm device is created. Maybe dmytro you could have a try with this?

tzhang449: Agree, I will try cache the BAR value when pkvm attach ptdev and audit the BAR write based this.

arch/x86/kvm/vmx/pkvm/hyp/pci.c
tzhang449: So here I should use atomic test of shadow_vm_handle right?

semihalf-maluka-dmytro: I agree that ptdev->lock does not help here yet (and atomic read of ptdev->shadow_vm_handle would not help either), since the issue is not just how to obtain correct ptdev attach status but how to ensure that this status will not change while we try to write BAR (i.e. to prevent security issue with ptdev already attached and BAR write succeeded immediately after, or a correctness issue with ptdev already detached and BAR write failed immediately after).

So yeah, it looks like ptdev->lock should be held until BAR write finishes.

@JasonChenCJ,

>while currently we defer the setup of guest EPT mapping for ptdev BAR untill EPT violation.

It seems to me that we need to change that anyway, e.g. due to security reasons: some secure MMIO may need to be protected from the host even before the secure VM touches this MMIO page first time. So it seems we should do setup of guest EPT for BAR MMIO already at the VM startup (pKVM has all the needed information for that anyway). 

A draft idea of the final implementation:
- ptdev attach/detach is done together with the initial mapping or unmapping of BAR MMIO in the guest and host EPT, under ptdev->lock
- BAR write audit checks if ptdev is attached and if so, updates the guest and host EPT mappings correspondingly, also under ptdev->lock

Makes sense?

JasonChenCJ: >and atomic read of ptdev->shadow_vm_handle would not help either

yes, that's why I said it's a loose control here. Ideally, after ptdev attach, the BAR mmio should already be mapped to guest and BAR reprogramming from host is forbid. 

But meanwhile:
- For guest BAR mmio mapping, we rely host vfio do the emulation, we only know gpa for this BAR mmio when guest program vBAR. 
- For host BAR reprogramming, based on our ptdev attach timing, we still see(& need) BAR access from host after it. To be more detail for one important thing, based on @tzhang449 's analysis, the vfio pci device should be inited after pkvm_attach_ptdev (vfio_config_init is done during vfio_pci_core_enable, while vfio_config_init read BAR info for vdev-rbar[])

So, how about we do like below?
1. for BAR write, we hold ptdev->lock all the time (although it does not help for current solution, as we really get vdev->rbar[] after ptdev attach, and we have lock protection for pci cfg data read/write)
2. for BAR write, if the ptdev already assigned, we still allow BAR writing with previous value
3. continue rely vfio do the BAR emulation, which means the mmio EPT mapping of guest will be defer
4. regarding secure device which need remove MMIO page from host, I'd like to make it as a TODO, I suppose we should still allow host to reprogram its BAR config, but forbid its mmio access based on BAR programing value

arch/x86/kvm/vmx/pkvm/hyp/pci.c
tzhang449: Dmytro thanks for your feedback. I updated the patch to cache the BAR value when pkvm attach ptdev, and allow BAR write only if it is writing the cached value into the BAR. Looks like things are ok for now.

arch/x86/kvm/vmx/pkvm/hyp/pci.c
tzhang449: Updated the patch to release the lock after BAR write.

arch/x86/kvm/vmx/pkvm/hyp/pci.c
```diff
@@ -52,6 +52,47 @@ static int pci_mmcfg_write(u64 address, int size, unsigned long value)
...
+	union pci_cfg_addr_reg reg;
+
+	reg.enable = 1;
+	reg.bdf = bdf;
+	reg.reg = offset & (~0x3);
+
+	pci_cfg_space_write(&reg, offset & 0x3, size, value);
+}
+
+static bool host_vpci_cfg_space_audit_access(struct pkvm_ptdev *ptdev, u64 offset, int size, u32 value)
```
JasonChenCJ, line 80: prefer to use name host_vpci_cfg_data_allow_write.

arch/x86/kvm/vmx/pkvm/hyp/pci.c
```diff
@@ -106,15 +147,43 @@ static int host_vpci_cfg_addr_write(struct kvm_vcpu *vcpu, struct pkvm_pio_req *
...
+	if (ptdev) {
+		pkvm_spin_lock(&ptdev->lock);
+		if (!host_vpci_cfg_space_audit_access(ptdev, offset + req->port - PCI_CFG_DATA,
+			req->size, *req->value)) {
+			ret = -EINVAL;
+			goto out;
+		}
+	}
+
+	ret = pci_cfg_space_write(&host_vpci_cfg_addr, req->port - PCI_CFG_DATA, req->size, *req->value);
```
JasonChenCJ, line 168: we should cache the value again after allow writing.

arch/x86/kvm/vmx/pkvm/hyp/pci.c
tzhang449: Changed the name.

arch/x86/kvm/vmx/pkvm/hyp/pci.c
semihalf-maluka-dmytro: Thanks. I confirm that now it works ok on my side too.

semihalf-maluka-dmytro: There is still `/* Protects updates of pgt pointer */` comment in `ptdev.h`, it probably should be removed since now ptdev->lock is used not only for ptdev->pgt updates.

tzhang449: Agree, will remove this.

arch/x86/kvm/vmx/pkvm/hyp/io_emulate.c
```diff
@@ -191,3 +208,167 @@ int register_host_mmio_handler(unsigned long start, unsigned long end,
 
 	return ret;
 }
+
+/*
+ * mmcfg access in x86 only use simple mov instrcutions. So keep the decoder
+ * simple for now.
+ * TODO: make the decoder complete
```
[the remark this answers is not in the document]
tzhang449: Hi Dmytro, sorry for late reply. Here we do emulation because we don't want host to access MMCFG region directly. If we do PV, then host is still able to access the MMIO without pkvm audit. What do you think about it?

arch/x86/kvm/vmx/pkvm/hyp/pci.c
tzhang449: Removed the comment  in latest branch.

arch/x86/kvm/vmx/pkvm/hyp/io_emulate.c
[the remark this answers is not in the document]
JasonChenCJ: Seemed we need continue the old way by simple instruction decode, @semihalf-maluka-dmytro, if you agree, pls check if there is other comments for this series patch, if no, we will merge this PR.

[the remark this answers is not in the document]
tzhang449: Thanks for the explanation. Yeah, I think if we can use the PCI_Config opregion it would indeed be a cleaner approach. If we have to change the ACPICA code, it would be precarious.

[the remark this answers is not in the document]
semihalf-maluka-dmytro: Yep, I agree. The PR looks good to me.

semihalf-maluka-dmytro (approved):

PR #32, opened 2025-05-21
Fixes vcpu accessing race

cxdong:
Fixes two issues:
1) The race of the host to load the same vcpu on multiple pCPU simultaneously.
2) The host use PV interface to access the pkvm_vcpu which is not loaded on this CPU.

### DIFF
--- a/arch/x86/kvm/pkvm/pkvm.c
+++ b/arch/x86/kvm/pkvm/pkvm.c
@@ -495,34 +495,10 @@ static void unload_pkvm_vcpu(struct pkvm_vcpu *pkvm_vcpu)
 	put_pkvm_vcpu(pkvm_vcpu);
 }
 
-static void set_pkvm_vcpu_inuse(struct pkvm_vcpu *pkvm_vcpu)
-{
-	struct kvm_vcpu *vcpu = to_kvm_vcpu(pkvm_vcpu);
-	int cpu = raw_smp_processor_id();
-
-	if (vcpu->cpu != -1)
-		/* Already inuse */
-		return;
-
-	get_pkvm_vm(vcpu->kvm->arch.pkvm.pkvm_vm_handle);
-	vcpu->cpu = cpu;
-}
-
-static void set_pkvm_vcpu_free(struct pkvm_vcpu *pkvm_vcpu)
-{
-	struct kvm_vcpu *vcpu = to_kvm_vcpu(pkvm_vcpu);
-
-	if (vcpu->cpu != raw_smp_processor_id())
-		/* Not inuse in this CPU */
-		return;
-
-	vcpu->cpu = -1;
-	put_pkvm_vm(pkvm_vcpu->pkvm_vm);
-}
-
 static void pkvm_vcpu_load(struct pkvm_vcpu *pkvm_vcpu, int cpu)
 {
 	struct kvm_vcpu *vcpu;
+	int loaded_cpu;
 
 	if (WARN_ON_ONCE(!pkvm_vcpu))
 		return;
@@ -531,35 +507,35 @@ static void pkvm_vcpu_load(struct pkvm_vcpu *pkvm_vcpu, int cpu)
 		return;
 
 	vcpu = to_kvm_vcpu(pkvm_vcpu);
-
-	/*
-	 * Loading vcpu on a new CPU without putting it from the previous one
-	 * is not supported.
-	 */
-	if (WARN_ON_ONCE(vcpu->cpu != -1 && vcpu->cpu != cpu))
+	loaded_cpu = cmpxchg(&vcpu->cpu, -1, cpu);
+	if (loaded_cpu == -1) {
+		/* Not loaded yet */
+		get_pkvm_vm(vcpu->kvm->arch.pkvm.pkvm_vm_handle);
+	} else if (WARN_ON_ONCE(loaded_cpu != cpu)) {
+		/* Already loaded on another CPU */
 		return;
+	}
 
 	kvm_x86_call(vcpu_load)(vcpu, cpu);
-
-	set_pkvm_vcpu_inuse(pkvm_vcpu);
 }
 
 static void pkvm_vcpu_put(struct pkvm_vcpu *pkvm_vcpu)
 {
+	int cpu = raw_smp_processor_id(), loaded_cpu;
 	struct kvm_vcpu *vcpu;
 
 	if (WARN_ON_ONCE(!pkvm_vcpu))
 		return;
 
 	vcpu = to_kvm_vcpu(pkvm_vcpu);
-
+	loaded_cpu = cmpxchg(&vcpu->cpu, cpu, -1);
 	/* Only the CPU which has loaded this vcpu can do a put */
-	if (WARN_ON_ONCE(vcpu->cpu != raw_smp_processor_id()))
+	if (WARN_ON_ONCE(loaded_cpu != cpu))
 		return;
 
 	kvm_x86_call(vcpu_put)(vcpu);
 
-	set_pkvm_vcpu_free(pkvm_vcpu);
+	put_pkvm_vm(pkvm_vcpu->pkvm_vm);
 }
 
 static void pkvm_vcpu_update_state_from_host(struct pkvm_vcpu *pkvm_vcpu)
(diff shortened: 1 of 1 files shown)


arch/x86/kvm/pkvm/pkvm.c
```diff
@@ -495,34 +495,22 @@ static void unload_pkvm_vcpu(struct pkvm_vcpu *pkvm_vcpu)
 	put_pkvm_vcpu(pkvm_vcpu);
 }
 
-static void set_pkvm_vcpu_inuse(struct pkvm_vcpu *pkvm_vcpu)
+static inline void set_pkvm_vcpu_inuse(struct pkvm_vcpu *pkvm_vcpu)
```
maluka-dmytro: The compiler most certainly will inline it anyway (since it is called in just one place)?

Also we can just call get_pkvm_vm()/put_pkvm_vm() in pkvm_vcpu_load()/pkvm_vcpu_put() instead of using these wrappers?

maluka-dmytro: Also about the commit message: s/automatically/atomically/ ?

arch/x86/kvm/pkvm/pkvm.c
```diff
@@ -475,14 +475,24 @@ static struct pkvm_vcpu *load_pkvm_vcpu(struct kvm_vcpu *shared_vcpu, unsigned l
...
-		return NULL;
+	if (!is_kvm_vcpu_accessible(vcpu, fn))
+		goto out;
+
+	/*
+	 * Except the vcpu_load PV interface, the other vcpu accessing PV
+	 * interfaces requires the vcpu to be loaded on this CPU.
+	 */
+	if (vcpu->cpu != raw_smp_processor_id()) {
+		if (vcpu->cpu != -1 || fn != __pkvm__vcpu_load)
```
maluka-dmytro: If we allow `vcpu->cpu == -1` (i.e. allow the host to do `vcpu_run` and other PV ops without doing `vcpu_load` first), it means the host can do some PV ops for the same vCPU on two different pCPUs simultaneously (as long as it hasn't vcpu_load'ed this vCPU on any pCPU), causing races in pKVM?

Why exactly do we need to allow `vcpu->cpu == -1`?

maluka-dmytro: Also nit: this can be in one `if` instead of two?

arch/x86/kvm/pkvm/pkvm.c
cxdong: Sure, we can just call get_pkvm_vm/put_kvm_vm. I will update the commit message as well.

arch/x86/kvm/pkvm/pkvm.c
cxdong: > If we allow `vcpu->cpu == -1` (i.e. allow the host to do `vcpu_run` and other PV ops without doing `vcpu_load` first), it means the host can do some PV ops for the same vCPU on two different pCPUs simultaneously (as long as it hasn't vcpu_load'ed this vCPU on any pCPU), causing races in pKVM?

Let me try to understand this scenario better, the scenario is the host use vcpu_run PV interface to run a vcpu which is not loaded first, is this correct? If so, then vcpu->cpu == -1, so the 2nd if will check fn. As it is __pkvm_vcpu_run PV interface, not __pkvm_vcpu_load, so the 2nd if is true, and goes out, which returns NULL. This seems fine. No?

> 
> Why exactly do we need to allow `vcpu->cpu == -1`?

To allow the __pkvm_vcpu_load PV interface. This is the only PV interface so far has vcpu->cpu == -1.
So we only allow the host to do __pkvm_vcpu_load for the same vcpu on two different pCPUs simultaneously. And such race will be prevented by the other patch via the cmpxchg.

cxdong: > Also nit: this can be in one `if` instead of two?

Yes, it can be in one `if`, just a bit longer line.

```suggestion
        if (vcpu->cpu != raw_smp_processor_id() &&
            (vcpu->cpu != -1 || fn != __pkvm__vcpu_load))
                goto out;

```

maluka-dmytro: Yeah, I got confused, I thought this condition allows the host to run a vcpu  which is not loaded, but now I see it doesn't allow that. It would be easier for me to get that if it was written as `!(vcpu->cpu == -1 && fn == __pkvm__vcpu_load)` (to clearly emphasize that __pkvm__vcpu_load is the special case when cpu == -1 is allowed, while in all other cases it is not). :)

maluka-dmytro: BTW also just a thought: maybe it would be cleaner to not call `pkvm_x86_call(switch_to_guest_vcpu)(vcpu)` if fn is __pkvm__vcpu_load?

If so, then in in vmx_switch_to_guest_vcpu() we could remove the check `vmx->loaded_vmcs->cpu == -1`, and then the check `WARN_ON_ONCE(vmx->loaded_vmcs->cpu != cpu)` would cover more cases?

cxdong: > Yeah, I got confused, I thought this condition allows the host to run a vcpu which is not loaded, but now I see it doesn't allow that. It would be easier for me to get that if it was written as `!(vcpu->cpu == -1 && fn == __pkvm__vcpu_load)` (to clearly emphasize that __pkvm__vcpu_load is the special case when cpu == -1 is allowed, while in all other cases it is not). :)

Sure, I can update to make it easier to read.

cxdong: > BTW also just a thought: maybe it would be cleaner to not call `pkvm_x86_call(switch_to_guest_vcpu)(vcpu)` if fn is __pkvm__vcpu_load?
> 
> If so, then in in vmx_switch_to_guest_vcpu() we could remove the check `vmx->loaded_vmcs->cpu == -1`, and then the check `WARN_ON_ONCE(vmx->loaded_vmcs->cpu != cpu)` would cover more cases?

I initially want to make the `pkvm_x86_call(switch_to_guest_vcpu)(vcpu)` and `pkvm_x86_call(switch_to_host_vcpu)(vcpu)` in pair for easier reading. As the `pkvm_x86_call(switch_to_host_vcpu)(vcpu)` is called unconditionally, also made the `pkvm_x86_call(switch_to_guest_vcpu)(vcpu)` unconditionally. How do you think?

maluka-dmytro: > As the `pkvm_x86_call(switch_to_host_vcpu)(vcpu)` is called unconditionally, also made the `pkvm_x86_call(switch_to_guest_vcpu)(vcpu)` unconditionally.

Good point. So there are pros and cons. Ok, I guess we can keep it as is.

maluka-dmytro: > As the `pkvm_x86_call(switch_to_host_vcpu)(vcpu)` is called unconditionally, also made the `pkvm_x86_call(switch_to_guest_vcpu)(vcpu)` unconditionally.

Good point. So there are pros and cons. Ok, I guess we can keep it as is.

arch/x86/kvm/pkvm/pkvm.c
```diff
@@ -531,35 +507,34 @@ static void pkvm_vcpu_load(struct pkvm_vcpu *pkvm_vcpu, int cpu)
...
-	 * Loading vcpu on a new CPU without putting it from the previous one
-	 * is not supported.
-	 */
-	if (WARN_ON_ONCE(vcpu->cpu != -1 && vcpu->cpu != cpu))
+	loaded_cpu = cmpxchg(&vcpu->cpu, -1, cpu);
+	if (loaded_cpu == -1)
+		/* Not loaded yet */
+		get_pkvm_vm(vcpu->kvm->arch.pkvm.pkvm_vm_handle);
+	else if (WARN_ON_ONCE(loaded_cpu != cpu))
+		/* Already loaded on another CPU */
```
maluka-dmytro, line 524: nit: I suppose the recommended style is to add {} braces in such cases (i.e. when we have a comment above the statement).

cxdong: checkpath doesn't report coding style issues. And seems the Documentation/process/coding-style.rst doesn't explicitly mention the case for one line comment and another line of code, but this case sounds like to match the "multiple statements" mentioned in the doc?

Adding braces is fine to me. Will update the patch.

maluka-dmytro (approved):

PR #44, opened 2025-11-12
Pkvm v6.18 host trace

cxdong:
Add pkvm trace support for host vmexits.


arch/x86/include/asm/kvm_pkvm.h
```diff
@@ -50,34 +50,34 @@ struct pkvm_mem_info {
 	u64 prot;
 };
```
maluka-dmytro, line 51: nit: commit message: s/SQUISHME/SQUASHME/ ?

arch/x86/include/asm/kvm_pkvm.h
```diff
@@ -50,10 +50,13 @@ struct pkvm_mem_info {
...
 
+#define TO_PKVM_HC(f)		CONCATENATE(__pkvm__, f)
+
 enum pkvm_hc {
-	__pkvm__init_finalize,
-	__pkvm__enable_vmexit_trace,
-	__pkvm__dump_vmexit_trace,
+	#define PKVM_HC(f)	TO_PKVM_HC(f),
+	#include <asm/pkvm_hypercalls.h>
```
maluka-dmytro, line 57: nit: commit message: s/pkvm_hypercall/pkvm_hypercalls/ ?

arch/x86/include/asm/pkvm_trace.h
```diff
@@ -14,12 +14,14 @@ struct vmexit_stats {
 
 struct perf_data {
 	struct vmexit_stats vmexit_reasons[MAX_EXIT_REASONS];
+	struct vmexit_stats hypercalls[MAX_EXIT_REASONS];
```
maluka-dmytro: MAX_PKVM_HYPERCALLS ?

arch/x86/kvm/pkvm/mmu.c
```diff
@@ -612,3 +612,124 @@ int pkvm_hyp_donate_host_mmio_locked(unsigned long phys, unsigned long size)
 
 	return pkvm_pgtable_map(&host_mmu, phys, phys, size, prot);
 }
+
+/**
+ * pkvm_host_share_hyp() - Shares pages from host to hypervisor.
```
maluka-dmytro: nit: "Share host pages with the hypervisor."?

And similarly for pkvm_host_unshare_hyp(): "Un-share host pages with the hypervisor."?

arch/x86/kvm/pkvm/mmu.c
```diff
@@ -612,3 +612,124 @@ int pkvm_hyp_donate_host_mmio_locked(unsigned long phys, unsigned long size)
...
+		page = pkvm_phys_to_page(phys);
+
+		/*
+		 * Even if host_share_hyp_count is 0 because the page is
+		 * shared with a guest, not with the hypervisor, it still
+		 * means a pKVM bug, since pkvm_host_unshare_hyp() is only
+		 * used by pKVM itself, for pages that are known to be
+		 * shared with the hypervisor.
+		 */
+		BUG_ON(!page->host_share_hyp_count);
```
maluka-dmytro, line 724: Just a note: I suggested this BUG_ON myself, but actually we could go ahead and implement a logic similar to the logic in pkvm_host_share_hyp(), i.e. first check page states and host_share_hyp_count counters for all pages, and then (if check succeeded) update page states and counters in a separate loop, to avoid this BUG_ON and thus make pkvm_host_unshare_hyp() more robust (and more consistent: in case of check_host_mem_pgstate() failure above we return an error instead of BUG_ON, so for consistency we could return an error in this case as well?).

But I'm probably fine either way.

arch/x86/include/asm/pkvm_trace.h
```diff
@@ -0,0 +1,26 @@
+/* SPDX-License-Identifier: GPL-2.0 */
+#ifndef _ASM_X86_PKVM_TRACE_H
+#define _ASM_X86_PKVM_TRACE_H
+
+#include <asm/pkvm_spinlock.h>
+#include <asm/vmx.h>
+
+#define MAX_EXIT_REASONS	MAX_VMX_EXIT_REASONS
```
maluka-dmytro, line 13: So we have a VMX-specific implementation in the generic part of pKVM?

Well, I have no suggestion how to easily address this, so maybe let's leave it as is for now...

arch/x86/kvm/pkvm/trace.h
```diff
@@ -0,0 +1,11 @@
+/* SPDX-License-Identifier: GPL-2.0 */
+#ifndef _HYP_TRACE_H_
+#define _HYP_TRACE_H_
+
+#include <linux/kvm_host.h>
+
+void trace_vmexit_start(struct kvm_vcpu *vcpu);
+void trace_vmexit_end(struct kvm_vcpu *vcpu, u32 index);
```
maluka-dmytro: Maybe `pkvm_trace_vmexit_start()` and `pkvm_trace_vmexit_end()`?

arch/x86/kvm/pkvm.c
```diff
@@ -81,3 +82,11 @@ void __init pkvm_reserve(void)
 	kvm_info("Reserved %lld MiB at 0x%llx for pkvm\n", pkvm_mem_size >> 20,
 		 pkvm_mem_base);
 }
+
+void pkvm_init_debugfs(void)
```
maluka-dmytro: Maybe let's move the debugfs stuff to a separate file pkvm_debugfs.c like in v6.12? It is kind of an isolated functionality... And for instance, KVM also has a separate file debugfs.c

arch/x86/include/asm/kvm_pkvm.h
```diff
@@ -50,10 +50,13 @@ struct pkvm_mem_info {
...
+
+	MAX_PKVM_HYPERCALLS,
```
maluka-dmytro, line 59: nit: maybe add MAX_PKVM_HYPERCALLS in the next patch, which actually uses it? (In my v6.12 patchset I've already done that today.)

arch/x86/kvm/pkvm.c
```diff
@@ -105,6 +106,125 @@ static int enable_vmexit_trace(void *data, u64 val)
...
+			   prefix, get_vmexit_reason(i), stats->count, stats->cycles,
+			   stats->cycles / stats->count);
+
+		if (summary_stats) {
+			summary_stats->count += stats->count;
+			summary_stats->cycles += stats->cycles;
+		}
+
+		if (need_resched())
+			cond_resched();
```
maluka-dmytro: I'm wondering why we actually need this?

arch/x86/kvm/pkvm.c
```diff
@@ -105,6 +106,125 @@ static int enable_vmexit_trace(void *data, u64 val)
...
+	kfree(prefix);
+
+	return 0;
+}
+
+static int pkvm_dump_vmexit_trace(struct seq_file *m, struct perf_data *dump,
+				  unsigned long size)
+{
+	struct perf_data *summary = kmalloc(sizeof(struct perf_data), GFP_KERNEL_ACCOUNT);
+	struct perf_data *pervcpu = dump;
```
maluka-dmytro: nit: for consistency we might do `for (cpu = 0, pervcpu = dump;` in the loop below, instead of initializing it here?

maluka-dmytro:
When compiling this, I'm getting:

```
arch/x86/kvm/vmx/pkvm_init.c:777:19: error: conflicting types for ‘pkvm_init_finalize’; have ‘int(void)’
  777 | static __init int pkvm_init_finalize(void)
      |                   ^~~~~~~~~~~~~~~~~~
In file included from arch/x86/kvm/pkvm/mmu.h:6,
                 from arch/x86/kvm/vmx/common.h:8,
                 from arch/x86/kvm/vmx/vmx.h:21,
                 from arch/x86/kvm/vmx/pkvm_init.c:8:
arch/x86/kvm/pkvm/init_finalize.h:28:5: note: previous declaration of ‘pkvm_init_finalize’ with type ‘int(struct pkvm_mem_info *, int,  struct pkvm_init_ops *)’
   28 | int pkvm_init_finalize(struct pkvm_mem_info infos[], int nr_info,
      |     ^~~~~~~~~~~~~~~~~~
```

So `#include "mmu.h"` in vmx/common.h resolves to pkvm/mmu.h instead of mmu.h. I have no idea why...

I've worked around this by changing it to `#include "../mmu.h"`

arch/x86/include/asm/kvm_pkvm.h
cxdong: Yeah, will change that. But it is not so important as this patch will be squashed with some merged patches when merging this PR.

arch/x86/kvm/pkvm/mmu.c
cxdong: My understanding is that, for the `pkvm_host_share_hyp()`, we prefer the logic to first check page states and `host_share_hyp_count` for all pages in a for loop is because `pkvm_host_share_hyp()` is taking the input address from the host which is untrusted. So it is something can happen even there is no code bug at all in the pKVM.

For `pkvm_host_unshare_hyp()`, I cannot imagine if there is any case for this function to take the input address from the host. If so, it only takes the input address from the pKVM hypervisor, which is trusted. Then it is something shouldn't happen if we have no bug. Adding a similar logic to check `host_share_hyp_count` for unshare seems almost not quite useful as most likely we don't have such bug?

To me, this BUG_ON takes the same effect with the BUG_ON in the first for loop of the `pkvm_host_share_hyp()`, which is purely to make sure we don't have any code bug in the pKVM hypervisor.

From this point of view, I prefer we keep to use the BUG_ON to keep it simpler.

arch/x86/include/asm/pkvm_trace.h
cxdong: Probably it is fine because this vmx.h is in asm folder, which can be used by a generic header file, e.g., arch/x86/kvm/trace.h includes both asm/vmx.h and asm/svm.h to get different vmexit reasons.

arch/x86/kvm/pkvm/trace.h
cxdong: Sounds good to me.

arch/x86/kvm/pkvm.c
cxdong: Makes sense. Will move it to pkvm_debugfs.c

arch/x86/include/asm/kvm_pkvm.h
cxdong: Sounds good. Will do that.

arch/x86/kvm/pkvm.c
cxdong: This is to prevent this function running too long to block other tasks in a non-preempted kernel. Although we know the loop number (MAX_EXIT_REASONS + MAX_PKVM_HYPERCALLS), but this is for each CPU, and later will be for each VM and each VM's each vCPU. So seems it is safe to have a need_resched() check. Maybe this can be put to the place which calls the `print_perf_data` rather than here, as most likely it is not necessary to reschedule just for the (MAX_EXIT_REASONS + MAX_PKVM_HYPERCALLS) number of loops, so that we don't need to check this so frequently.

cxdong:
>So #include "mmu.h" in vmx/common.h resolves to pkvm/mmu.h instead of mmu.h. I have no idea why...
I've worked around this by changing it to #include "../mmu.h"

What is the -I compiling options for pkvm_init.c? From my side, it is "-I ../arch/x86/kvm -Werror -I../arch/x86/kvm -Iarch/x86/kvm" (got this from .pkvm_init.o.cmd file), and it doesn't include arch/x86/kvm/pkvm/ folder, so will not resolve to pkvm/mmu.h. It makes sense as pkvm_init.c is not part of the pKVM hypervisor code.

Would you mind share the compiling instructions so that I can try this from my side?

arch/x86/kvm/pkvm/mmu.c
maluka-dmytro: Yeah, from this point of view it makes sense.

And this actually makes me think: would it be even better to add BUG_ON for the above check_host_mem_pgstate() check as well, and for `size == 0`, and make pkvm_host_unshare_hyp() void?

arch/x86/include/asm/pkvm_trace.h
maluka-dmytro: Yeah, it should still compile even if CONFIG_KVM_INTEL is disabled. But my point is that, e.g. in a hypothetical pKVM-AMD implementation, this vmexit tracing implementation will be still enabled, and will produce meaningless results (since the values of MAX_EXIT_REASONS and EXIT_REASON_VMCALL have nothing to do with SVM vmexits)?

maluka-dmytro: I have some vague ideas how we could refactor this: e.g. host_vmx.c could provide vendor callbacks like vmexit_reason_count(), vmexit_reason_to_index() and vmexit_reason_is_hypercall()...

But maybe for now, to save the effort, we may just add a comment like: "TODO: the implementation of vmexit tracing is VMX-specific. Make it generic." ?

cxdong: > Yeah, it should still compile even if CONFIG_KVM_INTEL is disabled. But my point is that, e.g. in a hypothetical pKVM-AMD implementation, this vmexit tracing implementation will be still enabled, and will produce meaningless results (since the values of MAX_EXIT_REASONS and EXIT_REASON_VMCALL have nothing to do with SVM vmexits)?

The MAX_EXIT_REASONS can be defined as a MAX(MAX_VMX_EXIT_REASONS, MAX_SVM_EXIT_REASONS)? If they are both enabled, select the larger value as MAX_EXIT_REASONS. If any of them is not enabled, we can redefine the disabled one as 0, so that only the enabled one can be selected? Or even always select the larger value to simplify?

Regarding the vmexit reason name to print, may be the trace data should contain a flag which is set by the pKVM hypervisor to indicate which vendor it is, and then pkvm_debugfs.c show different vmexit reason names accordingly.

> But maybe for now, to save the effort, we may just add a comment like: "TODO: the implementation of vmexit tracing is VMX-specific. Make it generic." ?

Adding a TODO makes sense to me.

arch/x86/kvm/pkvm/mmu.c
cxdong: Yeah, also makes sense to me.

maluka-dmytro:
In my .pkvm_init.o.cmd I see:

> savedcmd_arch/x86/kvm/vmx/pkvm_init.o := gcc -Wp,-MMD,arch/x86/kvm/vmx/.pkvm_init.o.d -nostdinc -I./arch/x86/include -I./arch/x86/include/generated -I./include -I./include -I./arch/x86/include/uapi -I./arch/x86/include/generated/uapi -I./include/uapi -I./include/generated/uapi -include ./include/linux/compiler-version.h -include ./include/linux/kconfig.h -include ./include/linux/compiler_types.h -D__KERNEL__ -std=gnu11 -fshort-wchar -funsigned-char -fno-common -fno-PIE -fno-strict-aliasing -mno-sse -mno-mmx -mno-sse2 -mno-3dnow -mno-avx -fcf-protection=branch -fno-jump-tables -m64 -falign-jumps=1 -falign-loops=1 -mno-80387 -mno-fp-ret-in-387 -mpreferred-stack-boundary=3 -mskip-rax-setup -march=x86-64 -mtune=generic -mno-red-zone -mcmodel=kernel -mstack-protector-guard-reg=gs -mstack-protector-guard-symbol=__ref_stack_chk_guard -Wno-sign-compare -fno-asynchronous-unwind-tables -mindirect-branch=thunk-extern -mindirect-branch-register -mindirect-branch-cs-prefix -mfunction-return=thunk-extern -fno-jump-tables -mharden-sls=all -fpatchable-function-entry=16,16 -fno-delete-null-pointer-checks -O2 -fno-allow-store-data-races -fstack-protector-strong -fno-omit-frame-pointer -fno-optimize-sibling-calls -ftrivial-auto-var-init=zero -fno-stack-clash-protection -pg -mrecord-mcount -mfentry -DCC_USING_FENTRY -fmin-function-alignment=16 -fstrict-flex-arrays=3 -fno-strict-overflow -fno-stack-check -fconserve-stack -fno-builtin-wcslen -Wall -Wextra -Wundef -Werror=implicit-function-declaration -Werror=implicit-int -Werror=return-type -Werror=strict-prototypes -Wno-format-security -Wno-trigraphs -Wno-frame-address -Wno-address-of-packed-member -Wmissing-declarations -Wmissing-prototypes -Wframe-larger-than=2048 -Wno-main -Wno-dangling-pointer -Wvla-larger-than=1 -Wno-pointer-sign -Wcast-function-type -Wno-array-bounds -Wno-stringop-overflow -Wno-alloc-size-larger-than -Wimplicit-fallthrough=5 -Werror=date-time -Werror=incompatible-pointer-types -Werror=designated-init -Wenum-conversion -Wunused -Wno-unused-but-set-variable -Wno-unused-const-variable -Wno-packed-not-aligned -Wno-format-overflow -Wno-format-truncation -Wno-stringop-truncation -Wno-override-init -Wno-missing-field-initializers -Wno-type-limits -Wno-shift-negative-value -Wno-maybe-uninitialized -Wno-sign-compare -Wno-unused-parameter -g -I ./arch/x86/kvm -iquote arch/x86/kvm/pkvm    -DKBUILD_MODFILE='"arch/x86/kvm/pkvm-intel"' -DKBUILD_BASENAME='"pkvm_init"' -DKBUILD_MODNAME='"pkvm_intel"' -D__KBUILD_MODNAME=kmod_pkvm_intel -c -o arch/x86/kvm/vmx/pkvm_init.o arch/x86/kvm/vmx/pkvm_init.c  

So in particular there is:

```
-I ./arch/x86/kvm -iquote arch/x86/kvm/pkvm
```

The compiler is gcc 14.2.0.

arch/x86/kvm/pkvm.c
maluka-dmytro: Ok, makes sense. Just an ignorant question: any advantage of checking need_resched() instead of just calling cond_resched() unconditionally?

maluka-dmytro:
With clang 19.1.7, I still have the same problem, and still have `-I ./arch/x86/kvm -iquote arch/x86/kvm/pkvm` in .pkvm_init.o.cmd.

Attached my kernel config: [config-debian-trixie-6.18-rc3-pkvm.txt](https://github.com/user-attachments/files/23526504/config-debian-trixie-6.18-rc3-pkvm.txt)

arch/x86/kvm/pkvm.c
cxdong: Probably they are checking for different purpose. I didn't fully understand the code, but seems need_resched() is to check if the rescheduling is needed or not, and cond_resched() just check if the rescheduling can be performed safely or not, and if it is safe then rescheduling, but may not be necessary.

maluka-dmytro: Ok, SGTM anyway.

cxdong:
Thanks for the sharing! Now I understand why I didn't encounter this issue. It is because I was using `#make LLVM=1 O=output` thus the -iquote arch/x86/kvm/pkvm doesn't have a mmu.h, so it has to resolve the mmu.h from ../arch/x86/kvm. I made a fix and the compiling is passed from my side with `#make LLVM=1`

arch/x86/include/asm/pkvm_trace.h
maluka-dmytro: AFAICS the problem is that the maximum SVM vmexit reason is 0x8000ffff or something like that. :)

And actually VMX exit reasons values AFAICS are also not fully contiguous (so our perf_data consumes a bit more memory than needed).

So perhaps a clean unified solution for both VMX and SVM would be to use the VMX_EXIT_REASONS and SVM_EXIT_REASONS arrays to map a vmexit reason to an index in the perf_data->vmexit_reasons array, instead of directly using the reason itself as the index. (That would mean consuming a bit more CPU cycles to look up the *_EXIT_REASONS array during tracing, but they would only be consumed if tracing is enabled anyway).

BTW also I guess we could reuse KVM_ISA_VMX and KVM_ISA_SVM from trace.h to distinguish between VMX and SVM vmexits (e.g. vendor-specific pKVM code could pass `isa` to trace_vmexit_end()).

...But again, I guess for now it's ok to keep the current implementation.

maluka-dmytro (approved):

arch/x86/kvm/pkvm.c
cxdong: Did some further experiment on this. With adding some delay with interrupt disabled, both `need_resched()` and `cond_resched()` can return 1 (for `need_resched()` means requires rescheduling, and for `cond_resched()` means rescheduling was happened) after enabling the interrupt. Decreasing the delay can result in both of them returning 0 (for `need_resched()` means not requires rescheduling, and for `cond_resched()` means no rescheduling happened).

By tracing the code,  `set_tsk_need_resched()` is the helper to set rescheduling for `need_resched()`, and `set_preempt_need_resched()` is the helper to set rescheduling for `cond_resched()`, and these two helpers are mostly called together. E.g., `rcu_sched_clock_irq()`, `resched_curr()`. For `resched_curr()`, if it needs to reschedule on a remote CPU, it will not do `set_preempt_need_resched()` but send a rescheduling IPI and the target cpu will do `set_preempt_need_resched()` eventually. Based on this, just calling cond_resched() is good enough. The difference is `need_resched()` checks the flags in current and `cond_resched()` checks the percpu preempt_count, which is in a single cacheline and should be even faster.

cxdong: Hi @maluka-dmytro, I plan to merge this PR with calling cond_resched() directly. Suppose this is also your preference, as you ever asked about this, and the above investigation shows checking need_resched() is not necessary.

maluka-dmytro: Yeah, SGTM. Thanks for further looking into this.

PR #49, opened 2025-12-12
pkvm-v6.18 fixes and improvements

maluka-dmytro:
- Make pkvm_hyp_donate_host() void and let it BUG_ON in case of any error, similarly to pkvm_host_unshare_hyp(), to facilitate catching bugs with incorrect usage of pkvm_hyp_donate_host() in pKVM code.
- Fix one such bug: when IPIv is disabled, pid_table is not allocated and not donated to pKVM, but we still try to undonate it when destroying the VM.

### DIFF
--- a/arch/x86/kvm/pkvm/mmu.c
+++ b/arch/x86/kvm/pkvm/mmu.c
@@ -715,18 +715,18 @@ void pkvm_host_unshare_hyp(unsigned long phys, unsigned long size)
 {
 	unsigned long end = PAGE_ALIGN(phys + size);
 	struct pkvm_page *page;
+	int ret;
 
-	BUG_ON(size == 0);
+	if (size == 0) {
+		ret = -EINVAL;
+		goto out;
+	}
 
 	pkvm_host_mmu_lock();
 
-	/*
-	 * The input range [phys, phys + size) is from the pKVM hypervisor which
-	 * is a trusted source, and suppose the pKVM hypervisor will only unshare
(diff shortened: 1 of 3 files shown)


arch/x86/kvm/pkvm/mmu.c
```diff
@@ -573,13 +568,11 @@ int pkvm_hyp_donate_host(unsigned long phys, unsigned long size, bool clear)
...
 	 */
-	BUG_ON(ret = pkvm_pgtable_map(&host_mmu, phys, phys, size, prot));
+	BUG_ON(pkvm_pgtable_map(&host_mmu, phys, phys, size, prot));
 
 	set_host_mem_pgstate(phys, size, PKVM_PAGE_OWNED);
-unlock:
-	pkvm_host_mmu_unlock();
 
-	return ret;
+	pkvm_host_mmu_unlock();
```
cxdong, line 580: It seems each function call in `pkvm_hyp_donate_host` should be wrapped by `BUG_ON()`. Does it look better if we just use `BUG_ON(ret)` here?

maluka-dmytro: May be not a bad idea, yeah. Although it makes the code a bit longer, but OTOH it would make it easier for us to adjust this code if we decide to return this error in the future.

Pushed updated version with your suggestion. Also accordingly applied the same change to pkvm_host_unshare_hyp(), just to keep it consistent.

cxdong: Thanks for the fixes. I will squash these patches.

PR #50, opened 2025-12-21
pkvm-v6.18: Fix worst-case calculation in __pkvm_pgtable_max_pages()

maluka-dmytro:
Another small fix to pKVM-v6.18 branch.

### DIFF
--- a/arch/x86/include/asm/kvm_pkvm.h
+++ b/arch/x86/include/asm/kvm_pkvm.h
@@ -262,6 +262,16 @@ static inline unsigned long __pkvm_pgtable_max_pages(unsigned long nr_pages)
 		total += nr_pages;
 	}
 
+	/*
+	 * For each level except the last one, may need an extra page table
+	 * if the VA range is not aligned to the next level's page size.
+	 * For example, range [0x1ff000, 0x201000) consists of just two
+	 * 4K pages, however, not one but two page tables at the first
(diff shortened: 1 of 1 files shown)


arch/x86/include/asm/kvm_pkvm.h
```diff
@@ -262,6 +262,16 @@ static inline unsigned long __pkvm_pgtable_max_pages(unsigned long nr_pages)
...
+	/*
+	 * For each level except the last one, may need an extra page table
+	 * if the VA range is not aligned to the next level's page size.
+	 * For example, range [0x1ff000, 0x201000) consists of just two
+	 * 4K pages, however, not one but two page tables at the first
+	 * level are required for mapping this range, since it crosses the
+	 * 2M boundary.
+	 */
+	total += PKVM_PGTABLE_MAX_LEVELS - 1;
+
```
cxdong, line 274: I though if we can pass the VA range information to this function and calculate the required page table accurately, but seems this is more complicated than this and not necessary for the worst case calculation.

Thanks for the fix. Will squash this patch to pkvm-v6.18 branch.

maluka-dmytro: Yeah, I thought about the same too, but yeah, if we don't do that, it seems the worst thing that happens in practice is that we allocate just a few more extra unneeded pages.

PR #53, opened 2026-01-02
Reprivilege CPUs on pkvm initialization failure

vineethrp:


arch/x86/include/asm/pkvm_hypercalls.h
```diff
@@ -8,7 +8,7 @@ BUILD_BUG_ON(1)
 #endif
 
 /* Hypercalls used only during pKVM initialization */
-PKVM_HC(init_finalize)
+PKVM_HC(init)
```
cxdong, line 11: As I saw the discussion in the google chat that, we may squash or reorder some patches in this serial. If so, it seems this patch can be a `SQUASHME` patch? If so, I can squash it to pkvm-v6.18 after the review.

arch/x86/kvm/pkvm/init.c
```diff
@@ -250,28 +250,27 @@ static int finalize_global(struct pkvm_mem_info infos[], int nr_infos,
...
 {
 	hyp_mmu_finalize_fn_t hyp_mmu_finalize_fn = init_ops ? init_ops->hyp_mmu_finalize :
 							       NULL;
 	host_mmu_finalize_fn_t host_mmu_finalize_fn = init_ops ? init_ops->host_mmu_finalize :
 								 NULL;
-	static bool global_finalized;
+	static bool global_initialized;
 	int ret;
 
 	if (this_cpu_read(cpu_finalized))
```
cxdong: This per cpu variable is still called `cpu_finalized`. Shall we rename it as well?

arch/x86/kvm/vmx/pkvm_init.c
```diff
@@ -1172,29 +1230,27 @@ int __init vmx_pkvm_init(void)
...
+	if (ret)
 		goto out;
-	}
 
+	/*
+	 * Marking start of the hypervisor initialization phase. We need to
+	 * do a rollback operation as part of disabling pKVM if any failure
+	 * happens from now on.
+	 */
+	pkvm_hyp_init_started = true;
```
cxdong: Maybe the reason we need this variable is that we have only one rollback function `pkvm_hyp_init_failed` which does two things: one is to re-privilege each CPU and the other is to `init_finalize` with `false`? If this understanding is correct, seems we can remove this variable if we split this rollback functions into two: E.g., if `pkvm_host_deprivilege_cpus` is failed, we go to the rollback function e.g., `pkvm_host_reprivilege_cpus` which reprivileges each CPUs. If `pkvm_hyp_init` is failed, we go to e.g., `pkvm_hyp_uninit` which does the `init_fianlzie` with `false` and `pkvm_host_reprivilege_cpus`?

```
ret = pkvm_host_deprivilege_cpus(pkvm);
if (ret)
    goto reprivilege;

ret = pkvm_host_init();
if (ret)
    goto uninit;
......
uninit:
    pkvm_hyp_uninit();
reprivilege:
    pkvm_host_reprivilege_cpu();
```

arch/x86/kvm/vmx/pkvm_init.c
```diff
@@ -1105,6 +1105,64 @@ static __init int pkvm_hyp_init(void)
...
+		return;
+
+	/*
+	 * If finalise operation started, it may be that hypervisor would
+	 * have revoked access to memory that may be needed for a clean
+	 * rollback. Request hypervisor to undo the memory protection to
+	 * continue the rollback.
+	 */
+	if (pkvm_hyp_init_started)
+		pkvm_hypercall(init_finalize, false);
```
cxdong: As I understand it, the `init_fianlize` with `false` is to make sure the pKVM hypervisor used memory is re-mapped in the host EPT, and the reason is that the code below from the host side needs to access the `struct pkvm_hyp` to get the `vcpu->mode` to know whether we should reprivilege a CPU or not, is this correct?

If so, maybe we can simplify this by adding a static per-CPU variable that is owned by the host itself. The host can set this variable to true if this CPU is deprivileged so that we can know if we should reprivilege it here.

Another approach is to leverage the exception table. As I saw, the later patch is using the below code to send vmcall to reprivilege:

```
	asm volatile(
		"vmcall\n"
		"endbr64\n"
		: "=a"(ret)
		: "a"(__pkvm__reprivilege_cpu)
		: "memory");
```
If a CPU is not in non-root mode, executing a vmcall instruction will cause #UD. Then we can leverage the exception table to capture this #UD and jump to the exception handler we defined, E.g.,

```
          asm goto(
                  "1: vmcall\n"
                  "endbr64\n"
                  _ASM_EXTABLE(1b, %l[do_exception])
                  : "=a"(ret)
                  : "a"(__pkvm__reprivilege_cpu)
                  : "memory"
                  : do_exception);

  do_exception:
```

Either way is fine with me if they can help to avoid this `init_finalize` with `false`.

arch/x86/kvm/pkvm/vmx/host_repriv.c
```diff
@@ -0,0 +1,159 @@
...
+	tss = (tss_desc *)&guest_gdt[GDT_ENTRY_TSS];
+	tss->type = DESC_TSS;
+	asm volatile("ltr %w0"::"q" (GDT_ENTRY_TSS*8));
+
+	guest_cr0 = vmcs_readl(GUEST_CR0);
+	guest_cr3 = vmcs_readl(GUEST_CR3);
+	guest_cr4 = vmcs_readl(GUEST_CR4);
+	__pkvm_write_cr0(guest_cr0);
+	__pkvm_write_cr4(guest_cr4);
+	__pkvm_write_cr3(guest_cr3);
```
cxdong: Just wondering about the consideration of the current register restoring sequence? E.g., we load segment registers first, then CR registers.

arch/x86/kvm/pkvm/vmx/host_repriv.c
```diff
@@ -0,0 +1,159 @@
...
+}
+
+#define STRINGIFY_INNER(x) #x
+#define STRINGIFY(x) STRINGIFY_INNER(x)
```
cxdong: Maybe we can merge these two macros into one, as STRINGIFY is the only user of STRINGIFY_INNTER?

arch/x86/kvm/pkvm/vmx/host_repriv.c
```diff
@@ -0,0 +1,159 @@
...
+
+	guest_rsp = vmcs_readl(GUEST_RSP);
+	guest_ss = vmcs_read16(GUEST_SS_SELECTOR);
+	guest_cs = vmcs_read16(GUEST_CS_SELECTOR);
+	guest_rip = vmcs_readl(GUEST_RIP) + vmcs_read32(VM_EXIT_INSTRUCTION_LEN);
+	guest_rflags = vmcs_readl(GUEST_RFLAGS);
+
+	asm volatile(
+		/* Restore general purpose registers */
+		RESTORE_VCPU_REG(VCPU_RCX, rcx)
```
cxdong: nit: it seems we are not saving too much of the code by defining this macro? If so, maybe we can just replace this to `"mov VCPU_RCX(%%rdi), %%rcx\n"` so that we don't need to define macro RESTORE_VCPU_REG, STRINGIFY and STRINGIFY_INNER.

arch/x86/kvm/pkvm/vmx/host_repriv.c
```diff
@@ -0,0 +1,159 @@
...
+#define STRINGIFY_INNER(x) #x
+#define STRINGIFY(x) STRINGIFY_INNER(x)
+
+/*
+ * Restores register state from memory pointed by rdi
+ * offset: offset of register backup in memory
+ * dest_reg: register to be restored.
+ */
+#define RESTORE_VCPU_REG(offset, dest_reg) \
+	"mov " STRINGIFY(offset) "(%%rdi), %%" #dest_reg "\n"
```
cxdong: nit: As this definition seems quite specific to the asm code usage in the function `pkvm_vmx_reprivilege_cpu` but not generic, shall we just define this macro inside the function `pkvm_vmx_reprivilege_cpu` to reflect that, and undefine it after it is used?

arch/x86/kvm/pkvm/vmx/host_repriv.c
```diff
@@ -0,0 +1,159 @@
...
+ * Restores host cpu state and returns to host in VMX root mode.
+ */
+void pkvm_vmx_reprivilege_cpu(unsigned long *vcpu_regs)
+{
+	/*
+	 * We manipulate SP in assembly. So don't use
+	 * stack for the variables.
+	 */
+	static u64 guest_rip, guest_rsp, guest_rflags;
+	static u16 guest_cs, guest_ss;
```
cxdong: nit: if we push these to the stack before manipulating the rbp, then it seems we don't need to use static here?

arch/x86/kvm/pkvm/vmx/host_repriv.c
```diff
@@ -0,0 +1,159 @@
...
+
+	repriv_restore_selectors();
+	repriv_restore_special_regs();
```
cxdong: Besides these registers, it seems the below MSRs also need to be restored:
1. MSR_IA32_DEBUGCTLMSR
2. MSR_CORE_PERF_GLOBAL_CTRL

The VMCS field has stored the value of these two MSRs.

arch/x86/include/asm/pkvm_hypercalls.h
vineethrp: Sure, I shall add SQUASHME details in the next iteration

arch/x86/kvm/pkvm/init.c
vineethrp: Sure, will do

arch/x86/kvm/vmx/pkvm_init.c
vineethrp: Other than the above reason, I had couple of other forward looking stuff. We need to clear protected memory and undo iommu initialization if iommu initialization failed. In 6.12 we clear the protected mem in host itself when initialization failed. So I was planning for that and to have a single failure handler which does all this.

Your above suggestion seems simpler for now and I shall adapt this. Will worry about rmem and iommu as we implement that. Thanks for the suggestion

arch/x86/kvm/vmx/pkvm_init.c
vineethrp: I was thinking about this and even if we can workaround accessing the vcpu struct, we might still need the hypervisor to donate the text area back(map in host pagetable) as we are still executing hypervisor code while we switch CR3 as part of reprivilege. I haven't seen a crash(probably because of caches already populated and not invalidated), but I think thats the safe way to go.

If you feel initi_finalize with false is confusing, I shall change the bool argument to an enum `pkvm_init_status` which could be `PKVM_INIT_STATUS_SUCCESS` or `PKVM_INIT_STATUS_FAILED`. What do you think?

arch/x86/kvm/pkvm/vmx/host_repriv.c
vineethrp: Thanks for catching this, yeah I need to switch the order. I am doing a small refactor where I do all vmcs read in a single go before restoring these registers. Will fix this as part of that.

arch/x86/kvm/pkvm/vmx/host_repriv.c
vineethrp: I did this initially as stringizing operator doesn't expand the argument, but we were passing VCPU_* macros which should be expanded before stringizing. But `RESTORE_VCPU_REG` will do the expansion for us and just tested that it works. Will remove STRINGIFY_INNER

arch/x86/kvm/pkvm/vmx/host_repriv.c
vineethrp: Since this is internal to this file and not exported, I thought it would be more readable to have it outside with a small comment as it is now. I can move it inside if you feel strongly about it.

arch/x86/kvm/pkvm/vmx/host_repriv.c
vineethrp: I tried this but did not work on my lenovo minipc(i5-13500). It works on brya and fatcat, but my mini pc doesn't come up(needs a hard reset). It doesn't have a serial port to see where it crashes. Disassembly shows that it uses only rsp to offset the variable and I don't understand exactly why it should crash/hang. So I thought its better to have it as static to be safe .

arch/x86/kvm/pkvm/vmx/host_repriv.c
vineethrp: Ahh, thanks for pointing this out. will fix.

arch/x86/kvm/vmx/pkvm_init.c
cxdong: > we might still need the hypervisor to donate the text area back(map in host pagetable) as we are still executing hypervisor code while we switch CR3 as part of reprivilege.

Do you mean that, the pKVM hypervisor still executes its own text after it switches to the host's CR3? If so, I think we don't need the pKVM hypervisor to donate its memory area back to the host for this purpose. Donating back to the host means we map these memory pages in the host EPT, this is irrelevant with the host's CR3. The host CR3 always maps the pKVM hypervisor's data and text. So it should be safe for the pKVM hypervisor to execute its own text after it switches to the host's CR3 without donating its memory back to the host.

arch/x86/kvm/pkvm/vmx/host_repriv.c
cxdong: I am fine either way. Just a nit. So please feel free to keep it as it is now.

arch/x86/kvm/pkvm/vmx/host_repriv.c
cxdong: I just want to make sure that you mean the following code sequence doesn't work on the lenovo minipc but works on brya and fatcat, right?
```
		"pushq %0\n"
		"pushq %1\n"
		"pushq %2\n"
		"pushq %3\n"
		"pushq %4\n"

		/* Restore general purpose registers */
		RESTORE_VCPU_REG(VCPU_RCX, rcx)
		RESTORE_VCPU_REG(VCPU_RDX, rdx)
		RESTORE_VCPU_REG(VCPU_RBX, rbx)
		RESTORE_VCPU_REG(VCPU_RBP, rbp)
		RESTORE_VCPU_REG(VCPU_RSI, rsi)
		RESTORE_VCPU_REG(VCPU_R8, r8)
		RESTORE_VCPU_REG(VCPU_R9, r9)
		RESTORE_VCPU_REG(VCPU_R10, r10)
		RESTORE_VCPU_REG(VCPU_R11, r11)
		RESTORE_VCPU_REG(VCPU_R12, r12)
		RESTORE_VCPU_REG(VCPU_R13, r13)
		RESTORE_VCPU_REG(VCPU_R14, r14)
		RESTORE_VCPU_REG(VCPU_R15, r15)

		/* Restore RDI (last!) */
		RESTORE_VCPU_REG(VCPU_RDI, rdi)
```
If so, not sure if this issue can be reproduced if we run the pKVM hypervisor inside a QEMU VM on this lenovo minipc?

arch/x86/kvm/vmx/pkvm_init.c
```diff
@@ -1081,19 +1081,19 @@ static void do_pkvm_finalize(void *data)
...
 				 (unsigned long)pkvm_sym(pkvm_vmx_init_ops));
 
 	if (data)
 		*(int *)data = ret;
 }
 
-static __init int pkvm_init_finalize(void)
+static __init int pkvm_hyp_init(void)
 {
 	int ret, cpu, finalize_ret;
```
maluka-dmytro: `finalize_ret` should also be renamed?

arch/x86/kvm/vmx/pkvm_init.c
```diff
@@ -1081,19 +1081,19 @@ static void do_pkvm_finalize(void *data)
...
 
 	for_each_possible_cpu(cpu) {
-		ret = smp_call_function_single(cpu, do_pkvm_finalize,
+		ret = smp_call_function_single(cpu, do_pkvm_hyp_init,
 					       &finalize_ret, 1);
 		if (ret || finalize_ret) {
 			pr_err("Failed to finalize CPU%d: smp_call %d, finalize: %d\n",
```
maluka-dmytro: Error message text should also be updated?

arch/x86/kvm/pkvm/vmx/host_vmx.c
```diff
@@ -6,7 +6,7 @@
 #include "debug.h"
 #include "ept.h"
 #include "host_vmx.h"
-#include "init_finalize.h"
+#include "init.h"
```
maluka-dmytro: nit: while we're at it, `"pkvm/init.h"` would be more consistent? Since this header is in common code in arch/x86/kvm/pkvm/ (like the below `lapic.h` and `trace.h`), not in VMX-specific code in arch/x86/kvm/pkvm/vmx/ (like the above `ept.h` and `host_vmx.h`)?

arch/x86/kvm/pkvm/init.c
```diff
@@ -194,11 +194,11 @@ static int create_host_mmu(const struct pkvm_mem_info infos[], int nr_infos,
 }
 
 #define TMP_NR_INFOS	16
-static int finalize_global(struct pkvm_mem_info infos[], int nr_infos,
-			   struct pkvm_init_ops *init_ops)
+static int initialize_global(struct pkvm_mem_info infos[], int nr_infos,
```
maluka-dmytro: A comment in pkvm_host_ept_finalize() still refers to finalize_global().

arch/x86/kvm/pkvm/init.c
```diff
@@ -250,28 +250,27 @@ static int finalize_global(struct pkvm_mem_info infos[], int nr_infos,
 	if (ret)
 		return ret;
 
-	return hyp_g_finalize ? hyp_g_finalize() : 0;
+	return hyp_global_init ? hyp_global_init() : 0;
 }
 
-int pkvm_init_finalize(struct pkvm_mem_info infos[], int nr_infos,
-		       struct pkvm_init_ops *init_ops)
+int pkvm_init(struct pkvm_mem_info infos[], int nr_infos, struct pkvm_init_ops *init_ops)
```
maluka-dmytro: Comments in pkvm_lapic_send_init() still refer to CPUs "finalized".

arch/x86/kvm/pkvm/init.c.orig
```diff
@@ -0,0 +1,350 @@
+// SPDX-License-Identifier: GPL-2.0
```
maluka-dmytro: File accidentally added?

arch/x86/kvm/pkvm/lapic.c
```diff
@@ -61,9 +61,9 @@ void pkvm_lapic_send_init(int cpu)
 	/*
 	 * Pairs with the smp_store_release() in the setup_lapic().
 	 * If remote lapic is not ready, it means the remote CPU is not
-	 * finalized yet. In this case, it is not necessary to send INIT to kick
-	 * as this remote CPU will handle all the pending requests before being
-	 * finalized.
+	 * initialized yet. In this case, it is not necessary to send INIT
```
maluka-dmytro: Now this makes it sound quite ambiguous? What is "CPU is not initialized"? CPUs as such are all initialized by kernel even before pKVM initialization starts.

So, don't know, maybe just s/initialized/initialized by pkvm_init()/ ?

arch/x86/include/asm/pkvm_hypercalls.h
vineethrp: I have added SQUASHME, but not sure which patch to squash to(multiple patches touches the logic). So I could not add fixes tag to this patch. Sorry about this. Would you be able to figure out how to squash it?

arch/x86/kvm/pkvm/init.c
vineethrp: Done

arch/x86/kvm/vmx/pkvm_init.c
vineethrp: Done

arch/x86/kvm/vmx/pkvm_init.c
vineethrp: Ahh my bad, I confused EPT with host page table. I have implemented your suggestion 1 with percpu as it avoids an IPI and VMCALL

arch/x86/kvm/pkvm/vmx/host_repriv.c
vineethrp: Done.

arch/x86/kvm/pkvm/vmx/host_repriv.c
vineethrp: Done

arch/x86/kvm/pkvm/vmx/host_repriv.c
vineethrp: Yeah this is what I tried and system was not booting. Currently I refactored the code to do all vmcs reads in one place and before restoring registers. I store all the host data in a new struct and this needs the stack pushes to be done before rsi is updated(the struct fields is accessed by offsetting rsi). So I moved the iret stack population up in the code. I shall debug the minipc issue soon with qemu as you suggested

arch/x86/kvm/pkvm/vmx/host_repriv.c
vineethrp: Done

arch/x86/kvm/vmx/pkvm_init.c
vineethrp: done

arch/x86/kvm/vmx/pkvm_init.c
vineethrp: done

arch/x86/kvm/pkvm/init.c
vineethrp: done

arch/x86/kvm/pkvm/init.c
vineethrp: done

arch/x86/kvm/pkvm/init.c.orig
vineethrp: oops sorry, removed

arch/x86/kvm/pkvm/lapic.c
vineethrp: makes sense. Updated the comment

arch/x86/kvm/pkvm/init.c
```diff
@@ -299,3 +304,47 @@ int pkvm_init(struct pkvm_mem_info infos[], int nr_infos, struct pkvm_init_ops *
...
+	int cpu;
+
+	if (pkvm_initialized)
+		return -EPERM;
+
+	for_each_possible_cpu(cpu) {
+		if (!per_cpu(cpu_initialized, cpu))
+			return -EAGAIN;
+	}
+	pkvm_initialized = true;
```
maluka-dmytro: Rather nit: as we discussed a couple of months ago, it might be good to use WRITE_ONCE to set it, to guarantee that whenever the (already formally untrusted) host makes the reprivilege hypercall after that, pKVM sees the up-to-date value and the hypercall fails?

Though I guess this is an almost ridiculously theoretical issue (the host is not gonna become potentially compromised _that_ fast), so I guess it would be more for "documentation" purpose.

arch/x86/kvm/pkvm/init.c
```diff
@@ -299,3 +304,47 @@ int pkvm_init(struct pkvm_mem_info infos[], int nr_infos, struct pkvm_init_ops *
...
+{
+	if (pkvm_initialized)
+		return -EPERM;
+
+	if (init_ops && init_ops->reprivilege_cpu)
+		init_ops->reprivilege_cpu(vcpu->arch.regs);
+
+	/*
+	 * Reach here only if reprivilege operation is not defined
+	 * by vendor module or reprivilege fails.
```
maluka-dmytro: nit: we could return -EOPNOTSUPP if `init_ops && init_ops->reprivilege_cpu` is false, and -EFAULT only if reprivilege was actually tried?

arch/x86/kvm/vmx/pkvm_init.c
```diff
@@ -1105,6 +1104,25 @@ static __init int pkvm_hyp_init(void)
...
+	/* TODO: Implement reprivilege logic. */
+}
+
+static __init void pkvm_host_reprivilege_cpus(struct pkvm_hyp *pkvm)
+{
+	int cpu;
+
+	for_each_possible_cpu(cpu) {
+		struct kvm_vcpu *vcpu = pkvm->host_vcpus[cpu];
+		if (vcpu->mode == OUTSIDE_GUEST_MODE)
```
maluka-dmytro: Worth using READ_ONCE (like e.g. we do in pkvm_host_vmexit_main() when checking for pending requests)?

maluka-dmytro: Sorry, this is in the host, not in the hypervisor, so comparison with pkvm_host_vmexit_main() is a bit bogus...

Anyway, just noticed that in the next patches this check is replaced with checking the `deprivileged` per-CPU variable... BTW why not introduce that variable already in this patch?

Or, maybe this per-CPU variable is actually redundant, and we can use `vcpu->mode` instead?

vineethrp: Ahh sorry about that, I should have moved it to here. To use vcpu->mode, we would have to undonate the memory back to host and I had that logic before. But Chuanxiao suggested to do it this way instead to make the code simpler and revised it like this to accomodate his comment.

I shall move it to this patch in the next iteration. sorry about that.

arch/x86/kvm/pkvm/init.c
```diff
@@ -193,9 +193,14 @@ static int create_host_mmu(const struct pkvm_mem_info infos[], int nr_infos,
 	return 0;
 }
 
+/*
+ * pKVM's local copy of pkvm_init_ops. Set by the host before deprivilege
```
maluka-dmytro: > pKVM's local copy of pkvm_init_ops.

This sounds like `pkvm_init_ops` is a variable, and this is a copy of that variable. But `pkvm_init_ops` is not a variable name, it is a struct type name?

Also this sounds like this is a copy of a host's variable, but there is no such host's variable, there is only pKVM's own `pkvm_vmx_init_ops`?

...We can probably just remove this 1st sentence?

arch/x86/kvm/pkvm/vmx/vcpu_regs.h
```diff
@@ -0,0 +1,27 @@
...
+#define VCPU_RDI	(__VCPU_REGS_RDI * WORD_SIZE)
+
+#define VCPU_R8		(__VCPU_REGS_R8  * WORD_SIZE)
+#define VCPU_R9		(__VCPU_REGS_R9  * WORD_SIZE)
+#define VCPU_R10	(__VCPU_REGS_R10 * WORD_SIZE)
+#define VCPU_R11	(__VCPU_REGS_R11 * WORD_SIZE)
+#define VCPU_R12	(__VCPU_REGS_R12 * WORD_SIZE)
+#define VCPU_R13	(__VCPU_REGS_R13 * WORD_SIZE)
+#define VCPU_R14	(__VCPU_REGS_R14 * WORD_SIZE)
+#define VCPU_R15	(__VCPU_REGS_R15 * WORD_SIZE)
```
maluka-dmytro, line 24: Just a note: the same definitions are also in KVM's `arch/x86/kvm/vmx/vmenter.S` (which is BTW reused by pKVM as well, for guest VM entry). So in the future (for upstream etc) we may make this header not just pKVM-specific but used by `arch/x86/kvm/vmx/vmenter.S` as well.

arch/x86/kvm/vmx/pkvm_init.c
maluka-dmytro: Just a note: I've realized that using `vcpu->mode` may have actually been completely wrong: the `vcpu->mode` value is actually changed at runtime, by the pKVM hypervisor, as this value indicates whether the host vCPU is currently running, i.e. executing host code in VMX non-root (then it is IN_GUEST_MODE) or pKVM is handling a host vmexit (then it is OUTSIDE_GUEST_MODE). @cxdong am I correct?

arch/x86/kvm/pkvm/Makefile
```diff
@@ -39,7 +39,7 @@ pkvm-hyp-y			+= $(kernel-lib)/sort.o $(kernel-lib)/bsearch.o \
 kvm				:= ..
```
maluka-dmytro, line 39: nit: commit messages: `s/pKVM: x86: VMX:/pKVM: VMX:/` like we do in other patches? (`x86:` is redundant if there is `VMX:`?)

arch/x86/kvm/pkvm/vmx/host_repriv.c
```diff
@@ -0,0 +1,156 @@
...
+		: "m"(hcs->gdt), "m"(hcs->idt), "q"(GDT_ENTRY_TSS*8),
+		  "m"(hcs->ds), "m"(hcs->es), "m"(hcs->fs), "m"(hcs->gs)
+		: "memory"
+	);
+
+	wrmsrl(MSR_FS_BASE, hcs->fsbase);
+	wrmsrl(MSR_GS_BASE, hcs->gsbase);
+
+	wrmsrq(MSR_IA32_DEBUGCTLMSR, hcs->debugctl);
+	wrmsrq(MSR_CORE_PERF_GLOBAL_CTRL, hcs->perf_global_ctrl);
```
maluka-dmytro: Question: what about other MSRs stored in VMCS (EFER, IA32_CR_PAT, IA32_SYSENTER_CS etc)?

arch/x86/kvm/pkvm/vmx/host_repriv.c
```diff
@@ -0,0 +1,156 @@
...
+	tss->type = DESC_TSS;
+
+	__pkvm_write_cr4(hcs->cr4);
+	__pkvm_write_cr0(hcs->cr0);
+	__pkvm_write_cr3(hcs->cr3);
+
+	asm volatile (
+		"lgdt %0\n"
+		"lidt %1\n"
+		"ltr %w2\n"
```
maluka-dmytro, line 102: Question: what about LDT?

arch/x86/kvm/pkvm/init.c
vineethrp: makes sense. Done

arch/x86/kvm/pkvm/init.c
vineethrp: valid point. Done

arch/x86/kvm/pkvm/init.c
vineethrp: makes sense, done.

arch/x86/kvm/vmx/pkvm_init.c
```diff
@@ -1171,22 +1190,25 @@ int __init vmx_pkvm_init(void)
...
+
+	pkvm_hypercall(init_finalize);
 
 	pkvm_init_debugfs();
 
 	pr_info("Hypervisor is up and running!\n");
 	return 0;
+
+repriv_cpus:
+	pkvm_host_reprivilege_cpus();
```
maluka-dmytro, line 1279: So, how about temporarily making pKVM initialization _always_ fail (and thus trigger reprivileging), until pvVMCS is ready? So that: 1. we can enable pKVM without breaking KVM functionality. 2. we can easily test (teamfood) reprivileging functionality just by enabling pKVM?

Ideally we could do that as a part of the patch e636d62e4f8c which enables deprivileging (which means that reprivileging would need to be implemented even before that patch), so that there is no single commit that makes KVM non-workable. Although we probably don't need to make the order of patches so perfect for now.

arch/x86/kvm/pkvm/vmx/host_repriv.c
vineethrp: My understanding is - since we are in early boot and in the absence of userspave,  LDT is not used and could be ignored. @cxdong, is this assumption true?

arch/x86/kvm/pkvm/vmx/host_repriv.c
vineethrp: Good question, will defer to @cxdong :-)

arch/x86/kvm/pkvm/vmx/host_repriv.c
maluka-dmytro: Still, we do load it into the VMCS during deprivileging:
```
	/* Initialize LDTR */
	store_ldt(ldtr);
	vmcs_write16(GUEST_LDTR_SELECTOR, ldtr);
	vmcs_write32(GUEST_LDTR_AR_BYTES, 0x10000);
	vmcs_writel(GUEST_LDTR_BASE, 0x0);
	vmcs_write32(GUEST_LDTR_LIMIT, 0xffffffff);
```

so we assume it does matter?

Anyway, @cxdong probably knows more.

Side note: I'm curious where the above values 0x10000, 0x0, 0xffffffff come from.

arch/x86/kvm/pkvm/init.c
```diff
@@ -193,9 +193,13 @@ static int create_host_mmu(const struct pkvm_mem_info infos[], int nr_infos,
 	return 0;
 }
 
+/*
+ * Set by the host before deprivilege and used through the initialization process.
+ */
```
maluka-dmytro: nit: since this is a one-line comment now, it can be in one line?

arch/x86/include/asm/pkvm_hypercalls.h
cxdong: Sure, I can handle that.

arch/x86/kvm/vmx/pkvm_init.c
cxdong: @maluka-dmytro yes, you are correct. `vcpu->mode` can be modified by the pKVM hypervisor when handling a host vmexit. But the pKVM hypervisor will set it to IN_GUEST_MODE before entering the host. So if the host is running at this CPU, that means the vcpu is in non-root mode and `vcpu->mode` is `IN_GUEST_MODE` already.

Using `vcpu->mode` requires additional hypercalls to donate the pKVM's memory to the host, and using a host side varaible is eaiser and can avoid adding that.

arch/x86/kvm/pkvm/vmx/host_repriv.c
cxdong: Yeah, makes sense to restore them as well. EFER/PAT are the same for the both side but still makes sense to restore them in case these are modified by the pKVM hypervisor. Maybe the additiona list is:
MSR_IA32_SYSENTER_CS, MSR_IA32_SYSENTER_ESP, MSR_IA32_SYSENTER_EIP, MSR_EFER, MSR_IA32_CR_PAT

arch/x86/kvm/pkvm/vmx/host_repriv.c
cxdong: Yeah I have the same understanding, the ldtr is 0, and the AR/BASE/LIMIT could be a default value. E.g, 0x10000 for AR represents unusable. Maybe a more generic way is to read the AR/BASE/LIMIT from GDT if the ldtr is a non-zero value.

But maybe it makes more sense to restore the guest LDT selector as otherwise we will use the pKVM's LDT, although it is also 0.

arch/x86/kvm/pkvm/init.c
```diff
@@ -250,28 +250,27 @@ static int finalize_global(struct pkvm_mem_info infos[], int nr_infos,
...
+	if (!global_initialized) {
+		ret = initialize_global(infos, nr_infos, init_ops);
 		if (ret)
 			return ret;
 
-		global_finalized = true;
+		global_initialized = true;
 	} else {
 		/*
 		 * The pKVM hypervisor's MMU was already loaded on the first
```
cxdong, line 266: The comment below should also be updated? s/finalized CPU/initialized CPU/, s/global finalize/global initialization/?

arch/x86/kvm/pkvm/init.c
```diff
@@ -299,3 +303,46 @@ int pkvm_init(struct pkvm_mem_info infos[], int nr_infos, struct pkvm_init_ops *
...
+		if (!per_cpu(cpu_initialized, cpu))
+			return -EAGAIN;
+	}
+	WRITE_ONCE(pkvm_initialized, true);
+	if (init_ops)
+		init_ops->reprivilege_cpu = NULL;
+	/*
+	 * TODO: Move reprivilege logic to a separate
+	 * section and zero it out here.
+	 */
```
cxdong, line 331: Question: not so clear what this TODO is going to do?

arch/x86/kvm/pkvm/init.c
```diff
@@ -299,3 +303,46 @@ int pkvm_init(struct pkvm_mem_info infos[], int nr_infos, struct pkvm_init_ops *
 	this_cpu_write(cpu_initialized, true);
 	return 0;
 }
+
+/*
+ * Flag indicating if pkvm is initialized successfully.
+ * Used to enforce internal hypercalls to be unavailable
+ * for general use once pkvm is initialized.
+ */
+static bool pkvm_initialized __ro_after_init;
```
cxdong, line 298: Do we plan to use this flag at some other places? E.g., check if this flag is true or not before creating a guest VM? If this is not necessary, it seems we can remove this flag, as the `pkvm_reprivilege_vcpu()` will return negative value anyway if `!init_ops || !init_ops->reprivilege_cpu`, and `pkvm_init_finalize()` can just set `init_ops->reprivilege_cpu = NULL;` if each CPU's `cpu_initialized` is true. The drawback is that, `pkvm_init_finalize()` will not be a one time hypercall and it can be called multiple times, but calling it multiple times seems no side effact (if we don't consider the race of calling `pkvm_reprivilege_vcpu` and `pkvm_init_finalize` at the same time on different CPUs, which is not a real usage for the host).

arch/x86/kvm/pkvm/vmx/host_repriv.c
```diff
@@ -0,0 +1,156 @@
...
+
+	wrmsrl(MSR_FS_BASE, hcs->fsbase);
+	wrmsrl(MSR_GS_BASE, hcs->gsbase);
+
+	wrmsrq(MSR_IA32_DEBUGCTLMSR, hcs->debugctl);
+	wrmsrq(MSR_CORE_PERF_GLOBAL_CTRL, hcs->perf_global_ctrl);
+}
+
+/*
+ * Restores host cpu state and returns to host in VMX root mode.
```
cxdong: nit: Another place where a one-line comment can be in one line?

arch/x86/kvm/vmx/pkvm_init.c
```diff
@@ -1108,7 +1108,51 @@ static __init int pkvm_hyp_init(void)
...
+	 */
+	asm volatile(
+		"vmcall\n"
+		"endbr64\n"
+		: "=a"(ret)
+		: "a"(__pkvm__reprivilege_cpu)
+		: "memory");
+
+	/*
+	 * Switch back to RO GDT page
```
cxdong: nit: Another place where a one-line comment can be in one line?

arch/x86/kvm/pkvm/vmx/host_repriv.c
dmaluka: Ah, I recalled that, IIUC, the GUEST_LDTR_SELECTOR setting (or any other GUEST_xxx_SELECTOR setting) basically has no effect when entering a VM, since the hardware will use the "cached" values programmed via GUEST_xxx_BASE, GUEST_xxx_LIMIT, GUEST_xxx_AR_BYTES instead.

So we are not really preserving the host's LDT selector during deprivileging, i.e. we are not really assuming that LDT matters at this point. If that's ok, then it might be ok to not do anything with it during reprivileging either...

arch/x86/kvm/pkvm/vmx/host_repriv.c
dmaluka: > still makes sense to restore them in case these are modified by the pKVM hypervisor.

Or by the host? Actually, more likely by the host than by the pKVM hypervisor?

Thinking more about this: actually reprivileging happens quite soon after deprivileging, so it's probably quite unlikely that kernel will change the values of these MSRs between deprivileging and reprivileging. But still it's hard to be sure that it is _guaranteed_ that kernel will not change them. So it still seems safer to restore them as well.

arch/x86/kvm/pkvm/init.c
maluka-dmytro: > but calling it multiple times seems no side effact

For now. But I guess later we will extend pkvm_init_finalize() to do more stuff, like we do in v6.12 in pkvm_commit_finalise() (i.e. uninitialize IOMMU + restore host access to the pvmfw memory region so the host can clear it before starting reprivileging).

arch/x86/kvm/pkvm/init.c
maluka-dmytro: I recall we had an idea to clear the reprivileging code in memory after pKVM initialization, to reduce the risk of accidentally executing it and thus giving the untrusted host full privileges...

arch/x86/kvm/pkvm/init.c
vineethrp: yeah true.

arch/x86/kvm/pkvm/init.c
vineethrp: yeah correct. It might be simple to do, but I am not planning to get to it before pviommu effort. So I can remove this comment if it seems to be confusing.

arch/x86/kvm/pkvm/init.c
vineethrp: Done

arch/x86/kvm/pkvm/vmx/host_repriv.c
vineethrp: Thanks for the extra context. So, I think we can just ignore LDT for now.

arch/x86/kvm/pkvm/vmx/host_repriv.c
vineethrp: Implemented the restoring of above 4 MSRs.

arch/x86/kvm/pkvm/Makefile
vineethrp: Ahh I didn't realize the precedence. I have removed the redundant x86

arch/x86/kvm/pkvm/vmx/vcpu_regs.h
vineethrp: makes sense, will put a note on it on my side as well

arch/x86/kvm/pkvm/vmx/host_repriv.c
```diff
@@ -0,0 +1,166 @@
...
+		: "m"(hcs->gdt), "m"(hcs->idt), "q"(GDT_ENTRY_TSS*8),
+		  "m"(hcs->ds), "m"(hcs->es), "m"(hcs->fs), "m"(hcs->gs)
+		: "memory"
+	);
+
+	wrmsrl(MSR_FS_BASE, hcs->fsbase);
+	wrmsrl(MSR_GS_BASE, hcs->gsbase);
+
+	wrmsrq(MSR_IA32_DEBUGCTLMSR, hcs->debugctl);
+	wrmsrq(MSR_CORE_PERF_GLOBAL_CTRL, hcs->perf_global_ctrl);
```
maluka-dmytro: With such a testing hack:
```
--- a/arch/x86/kvm/vmx/pkvm_init.c
+++ b/arch/x86/kvm/vmx/pkvm_init.c
@@ -995,7 +995,7 @@ static __init void pkvm_host_deprivilege_cpu(void *data)
                goto vmxoff;
        }
 
-       ret = local_deprivilege_cpu();
+       ret = (cpu == 7) ? -EFAULT : local_deprivilege_cpu();
        if (ret) {
                pr_err("CPU%d deprivilege failed, ret %d\n", cpu, ret);
                goto vmxoff;
```

when running in qemu, it crashes during reprivileging:

```
[    0.891677] pkvm: Failed to deprivilege CPU7: smp_call 0, deprivilege: -14
[    0.893057] Oops: general protection fault, maybe for address 0x0: 0000 [#1] SMP NOPTI
[    0.894874] CPU: 0 UID: 0 PID: 0 Comm: swapper/0 Not tainted 6.18.0+ #178 PREEMPT(voluntary) 
[    0.896629] Hardware name: QEMU Standard PC (Q35 + ICH9, 2009), BIOS 1.16.3-debian-1.16.3-2 04/01/2014
[    0.896970] RIP: 0010:pkvm_vmx_reprivilege_cpu__pkvm+0x309/0x690
[    0.896970] Code: 0f 30 48 8b 05 68 53 bc 01 b9 d9 01 00 00 48 89 c2 48 c1 ea 20 0f 30 48 8b 05 5b 53 bc 01 b9 8f 03 00 00 48 89 c2 48 c1 ea 20 <0f> 30 48 8b 05 4e 53 bc 01 b9 74 01 00 00 48 89 c2 48 c1 ea 20 0f
[    0.896970] RSP: 0000:ffff88817a026f08 EFLAGS: 00010046
[    0.896970] RAX: 0000000000000000 RBX: ffff88817a02f120 RCX: 000000000000038f
[    0.896970] RDX: 0000000000000000 RSI: 0000000000000002 RDI: ffff88817a02f120
[    0.916488] RBP: ffff88817a026f18 R08: 0000000000000405 R09: 0000000000000000
[    0.916488] R10: 0000000000000000 R11: ffffc90000003ff8 R12: ffffffff835009bd
[    0.916488] R13: 0000000000000002 R14: ffff88817a02f000 R15: ffffffff83500970
[    0.916488] FS:  0000000000000000(0000) GS:ffff8881f5eaa000(0000) knlGS:0000000000000000
[    0.916488] CS:  0010 DS: 0000 ES: 0000 CR0: 0000000080050033
[    0.916488] CR2: ffff88817ffff000 CR3: 000000000323a001 CR4: 0000000000772ef0
[    0.916488] PKRU: 55555554
[    0.916488] Call Trace:
[    0.916488] Modules linked in:
[    0.916488] ---[ end trace 0000000000000000 ]---
[    0.916488] RIP: 0010:pkvm_vmx_reprivilege_cpu__pkvm+0x309/0x690
```

Debugging shows that it crashes on writing MSR_CORE_PERF_GLOBAL_CTRL. Apparently this MSR is not implemented by qemu. And I assume the hypervisor's extable is not workable at this point, since pkvm_init() has not even started.

On real hardware this MSR it implemented, so this crash is not observed.

Anyway, this trick fixes it:

```
--- a/arch/x86/kvm/pkvm/vmx/host_repriv.c
+++ b/arch/x86/kvm/pkvm/vmx/host_repriv.c
@@ -107,7 +107,8 @@ static inline void restore_host_special_regs(struct host_cpu_state *hcs)
        wrmsrl(MSR_GS_BASE, hcs->gsbase);
 
        wrmsrq(MSR_IA32_DEBUGCTLMSR, hcs->debugctl);
-       wrmsrq(MSR_CORE_PERF_GLOBAL_CTRL, hcs->perf_global_ctrl);
+       if (hcs->perf_global_ctrl)
+               wrmsrq(MSR_CORE_PERF_GLOBAL_CTRL, hcs->perf_global_ctrl);
        wrmsrq(MSR_IA32_SYSENTER_CS, hcs->sysenter_cs);
        wrmsrq(MSR_IA32_SYSENTER_ESP, hcs->sysenter_esp);
        wrmsrq(MSR_IA32_SYSENTER_EIP, hcs->sysenter_eip);
```

I suppose this fix is correct, since pKVM ensures that while running in the pKVM hypervisor, this MSR value is always 0 (see Kevin's patch 1604e0800c5b), so if it is 0 in the host as well, we don't need to restore it.

arch/x86/kvm/pkvm/init.c
cxdong: I see. Thanks for the explainations. Please keep this comment.

arch/x86/kvm/pkvm/vmx/host_repriv.c
vineethrp: Sorry about this, I haven't tested much on qemu after reviving my minipc for testing. Your fix looks good to me. The harder but correct way is to check pmu version id >= 2, but I guess this simple fix would be best as pkvm would not theoretically be supported on older hardware and this fix  be just for qemu

arch/x86/kvm/pkvm/init.c
cxdong: Got it. To prevent `pkvm_init_finalize()` from being called multiple times, `pkvm_init_finalize()` can also check `if (!init_ops || !init_ops->reprivilege_cpu)` to decide if continue or just return with 0. But I am also fine to keep this static variable which seems more straightforward.

arch/x86/kvm/pkvm/vmx/host_repriv.c
cxdong: Good cache. This reminds me that, the `wrmsrl` is going to deprecated, so it is better we can use `wrmsrq` instead.

And because MSR_CORE_PERF_GLOBAL_CTRL relies on some CPUID, and now the pKVM hypervisor can support the safe version of wrmsrq, this can also be fixed by using `wrmsrq_safe` rather than `wrmsrq`. To be noted, this should be done before we switching to the host's IDT and GS.

maluka-dmytro: > this can also be fixed by using `wrmsrq_safe` rather than `wrmsrq`. To be noted, this should be done before we switching to the host's IDT and GS.

The problem is that, for example, in this test with failing `local_deprivilege_cpu()`, reprivileging is done too early, when pKVM's exception fixup is not workable yet. I've just made an experiment: replaced `wrmsrq` with `wrmsrq_safe`, but it still crashes, the same way. I haven't investigated why exactly, but I suppose at least because pKVM hasn't called `sort_extable()` yet (since the host hasn't even made the `init` hypercall to pKVM yet)?

arch/x86/kvm/pkvm/vmx/host_repriv.c
cxdong: I feel this is irrelavent with the "cached". The "cached" thing makes the host VM itself to use the LDT base/limit/ar programeed by the VMCS rather than the value loaded from its own GDT memory page, which is the unusable hidden part we programmed in the VMCS before deprivilege.

But here after a CPU is deprivileged and vmexit to the pKVM hypervisor to reprivilege, the LDT selector and hiddent part (base/limit/AR) switch to the pKVM hypervisor's value(in real, the pKVM hypervisor doesn't set LDT so its LDT selector is always 0 and LDT hidden part is also unusable). If we don't restore the host's LDT in the root mode, then the LDT will be remained as the root mode's value. We are good without restore the host's LDT selector because both sides are not using LDT so the LDT selector and base/limit/ar are the same for them. But if the pKVM hypervisor uses LDT after a vmexit, the root mode may use a different LDT hidden part with the host, then this seems will be problematic without restoring the host's LDT (which is unused) during reprivileging?

arch/x86/kvm/pkvm/vmx/host_repriv.c
maluka-dmytro: Ok, just noticed that we indeed write MSRs after switching the IDT, not before (which explains why this #GP call trace is printed by the host kernel, i.e. why it is handled by the host's exception handler).

So tried this:

```
--- a/arch/x86/kvm/pkvm/vmx/host_repriv.c
+++ b/arch/x86/kvm/pkvm/vmx/host_repriv.c
@@ -88,6 +88,8 @@ static inline void restore_host_special_regs(struct host_cpu_state *hcs)
        __pkvm_write_cr0(hcs->cr0);
        __pkvm_write_cr3(hcs->cr3);
 
+       wrmsrq_safe(MSR_CORE_PERF_GLOBAL_CTRL, hcs->perf_global_ctrl);
+
        asm volatile (
                "lgdt %0\n"
                "lidt %1\n"
@@ -107,7 +109,6 @@ static inline void restore_host_special_regs(struct host_cpu_state *hcs)
        wrmsrl(MSR_GS_BASE, hcs->gsbase);
 
        wrmsrq(MSR_IA32_DEBUGCTLMSR, hcs->debugctl);
-       wrmsrq(MSR_CORE_PERF_GLOBAL_CTRL, hcs->perf_global_ctrl);
        wrmsrq(MSR_IA32_SYSENTER_CS, hcs->sysenter_cs);
        wrmsrq(MSR_IA32_SYSENTER_ESP, hcs->sysenter_esp);
        wrmsrq(MSR_IA32_SYSENTER_EIP, hcs->sysenter_eip);
```

and this helps: kernel is booting successfully.

But still, is it really guaranteed to work if pKVM hasn't called `sort_extable()` yet?

arch/x86/kvm/pkvm/vmx/host_repriv.c
maluka-dmytro: > But if the pKVM hypervisor uses LDT after a vmexit

Just wondering: do we foresee any use cases for using LDT in pKVM?

cxdong: > Just wondering: do we foresee any use cases for using LDT in pKVM?

No, I think we don't foresee any use cases for that.

arch/x86/kvm/pkvm/vmx/host_repriv.c
cxdong: > But still, is it really guaranteed to work if pKVM hasn't called sort_extable() yet?

This is a good point. It may not be safe without sorting. So your method should be more reliable.

maluka-dmytro: > The harder but correct way is to check pmu version id >= 2

Thinking more about correctness of my simple fix with `if (hcs->perf_global_ctrl)`... It may be not fully correct, but only because the existing pKVM's behavior seems not fully correct: pKVM always intercepts host's writes to MSR_CORE_PERF_GLOBAL_CTRL, and if the host writes a non-zero value to it, pKVM sets it in the host's VMCS field GUEST_IA32_PERF_GLOBAL_CTRL without checking if physical MSR_CORE_PERF_GLOBAL_CTRL is implemented. While pKVM should probably check that and inject #GP into the host if it is not implemented.

If we fix that, then my fix seems fully correct: it would be guaranteed that `hcs->perf_global_ctrl` never has a non-zero value if MSR_CORE_PERF_GLOBAL_CTRL is not implemented?

cxdong: Indeed. Although the real hardware has the MSR_CORE_PERF_GLOBAL_CTRL, a QEMU vm can be implemented not to support it. Created PR https://github.com/intel-staging/pKVM-IA/pull/55 to fix this.

Meanwhile, also moved the exception table sort from `pkvm_init_finalize` to the host side before deprivilege in PR https://github.com/intel-staging/pKVM-IA/pull/55. As excpetion fixup doesn't rely on the initialization done by the `pkvm_init_finalize`, sort the exception table can be done earlier. So move it to the deprivilege phase to make the exception fixup mechanism functional once the deprivileged is done. With this, we can also use `wrmsrq_safe` to fix this issue without checking `hcs->perf_global_ctrl`.

maluka-dmytro: Great, thanks. PR #55 looks good to me but I see it is not actually working as it should on real HW on hybrid CPU (i.e., on any CPUs we actually care about), I've added a comment in the PR about that. But that's hopefully easy to fix. 

With that fixed, I guess we should do `wrmsrq_safe(MSR_CORE_PERF_GLOBAL_CTRL, hcs->perf_global_ctrl)` before switching IDT, that seems cleaner than my fix.

Maybe we can also move writing other MSRs (maybe except FS_BASE and GS_BASE) before switching IDT, together with MSR_CORE_PERF_GLOBAL_CTRL. Or maybe no need to.

cxdong: PR https://github.com/intel-staging/pKVM-IA/pull/55 has been merged and squashed to pkvm-v6.18 branch

vineethrp: Great, thanks. I will rebase on that and fix the msr restore logic and update the PR soon

arch/x86/kvm/vmx/pkvm_init.c
vineethrp: I like this idea. I have added a separate "REVERTME" commit for this

arch/x86/kvm/pkvm/vmx/host_repriv.c
vineethrp: Moved all MSR restoration above GDT restoration and used wrmsrq_safe for MSR_CORE_PERF_GLOBAL_CTRL.

arch/x86/kvm/pkvm/init.c
```diff
@@ -282,6 +283,15 @@ int pkvm_init(struct pkvm_mem_info infos[], int nr_infos)
...
 		return ret;
 
+	/*
+	 * XXX: Revert
+	 * Temporrarily fail pkvm initialization until pVMCS is fully merged.
+	 * pKVM doesn't serve any real purpose until we have pVMCS ready and
+	 * this failure helps us test reprivilege code when pKVM is enabled.
+	 */
+	pkvm_err("%s: Explicitly triggering pkvm initialization failure!\n", __func__);
+	return -EFAULT;
```
maluka-dmytro: IIUC this is gonna fail after initializing the 1st CPU, so pKVM will not start initializing any other CPUs. So maybe it would make more sense to fail in the host in vmx_pkvm_init(), after returning from pkvm_hyp_init(), so that we still test full pKVM initialization path as well, before failing and reprivileging?

arch/x86/kvm/pkvm/init.c
```diff
@@ -282,6 +283,15 @@ int pkvm_init(struct pkvm_mem_info infos[], int nr_infos)
 	if (ret)
 		return ret;
 
+	/*
+	 * XXX: Revert
+	 * Temporrarily fail pkvm initialization until pVMCS is fully merged.
+	 * pKVM doesn't serve any real purpose until we have pVMCS ready and
+	 * this failure helps us test reprivilege code when pKVM is enabled.
```
maluka-dmytro: nit: besides this, this also allows enabling pKVM without breaking KVM functionality, by providing graceful fallback to regular KVM.

arch/x86/kvm/pkvm/init.c
vineethrp: I thought it should be okay as we do the full initialization on one cpu. But moving the failure to the host side after pkvm_hyp_init is fine with me. Changed the implementation.

arch/x86/kvm/pkvm/init.c
vineethrp: updated the comment

arch/x86/kvm/vmx/pkvm_init.c
cxdong: It seems `REORDER: pKVM: x86: Undo hypervisor initialization in case of failure` has done two things: one is to add the reprivilege PV interface in case of failure, another one is add init_finalize PV interface. Maybe it is better to split them into two patches so that they can be re-ordered independently, e.g., we can reorder the reprivilege PV interface patch before deprivileging, and reorder the init_finalize PV interface after the new init PV interface?

vineethrp: Makes sense. I have split the patch into multiple SQUASHME/REORDER patches now. Please have a look and let me know if this is okay

cxdong: Thanks for splitting this patch. They basically looks good to me.  Just the patch "SQUASHME: pkvm: VMX: Reprivilege host cpus if pkvm init fails" seems to be a REORDER patch but cannot be squashed in the https://github.com/intel-staging/pKVM-IA/commit/5dc05c1abda6e2d50dd0f6c070b87d7cc147ad84 because the reprivilege is a hypercall, and the commit https://github.com/intel-staging/pKVM-IA/commit/5dc05c1abda6e2d50dd0f6c070b87d7cc147ad84 doesn't have enable any hypercall yet. So it would be better if there is some commit message for this patch

vineethrp: Done, changed to REORDER and added verbose commit message.

arch/x86/kvm/pkvm/init.h
```diff
@@ -10,23 +10,28 @@ typedef int (*host_mmu_init_fn_t)(struct pkvm_pgtable *pgt, void *pool_base,
...
  *
- * @hyp_mmu_finalize:	Finalize the hypervisor mmu.
- * @host_mmu_init:	Initialize the host mmu.
- * @host_mmu_finalize:	Finalize the host mmu.
- * @hyp_global_init:	Initialize the hypervisor globally.
+ * @hyp_mmu_finalize:		Finalize the hypervisor mmu.
+ * @host_mmu_init:		Initialize the host mmu.
+ * @host_mmu_finalize:		Finalize the host mmu.
+ * @hyp_global_init:		Initialize the hypervisor globally.
+ * @reprivilege_cpu_fn_t:	Switch the cpu back to root mode. Called if deprivilege
```
maluka-dmytro: s/_fn_t// ?

vineethrp: Ahh sorry about that, fixed.

arch/x86/kvm/vmx/pkvm_init.c
cxdong: Thanks!

cxdong (approved):
This PR looks good to me now. @maluka-dmytro , wonderring if there is any more review comment from your side or any discussion not closed yet?

maluka-dmytro (approved):

maluka-dmytro:
Yep, the PR looks good to me.

I haven't had a chance to test it on ADL and PTL yet, but IIUC Vineeth did?

vineethrp:
> Yep, the PR looks good to me.
> 
> I haven't had a chance to test it on ADL and PTL yet, but IIUC Vineeth did?

I have tested on qemu, Raptorlake(minipc) and PTL and tried partial and full deprivilege failure scenarios and pkvm_hyp_init fail scenarios. All seems to work fine. Ran VMS as well on all these.  It looks okay with respect to all testing I have done

cxdong:
Thanks. I will merge this PR and squash/reorder these patches.

PR #54, opened 2026-01-04
Pkvm v6.18 helpers

maluka-dmytro:
Add a few convenient helpers, to reduce code duplication and improve code readability (especially with upcoming changes for guest MMU support, which will use these helpers too):

- `for_each_pkvm_page()` macro for iterating through `struct pkvm_page`s for all pages in the given memory range (same as pKVM-ARM's `for_each_hyp_page()`)
- `pkvm_pgtable_max_size()` returning max size of the VA space for the given page table
- `host_mmu_pte_prot()` returning the needed property bits for host MMU PTEs

### DIFF
--- a/arch/x86/kvm/pkvm/memory.h
+++ b/arch/x86/kvm/pkvm/memory.h
@@ -44,6 +44,12 @@ extern u64 __pkvm_vmemmap;
 #define pkvm_page_to_virt(page)	__pkvm_va(pkvm_page_to_phys(page))
 #define pkvm_page_to_pool(page)	(((struct pkvm_page *)page)->pool)
 
+/* Caution: __st is evaluated twice. */
+#define for_each_pkvm_page(__p, __st, __sz)						\
+	for (struct pkvm_page *__p = pkvm_phys_to_page(PAGE_ALIGN_DOWN(__st)),		\
+			      *__e = pkvm_phys_to_page(PAGE_ALIGN((__st) + (__sz)));	\
+	     __p < __e; __p++)
+
 /*
  * Refcounting for 'struct pkvm_page'.
  * pkvm_pool::lock must be held if atomic access to the refcount is required.
--- a/arch/x86/kvm/pkvm/mmu.c
+++ b/arch/x86/kvm/pkvm/mmu.c
@@ -607,8 +607,7 @@ out:
  */
 int pkvm_hyp_donate_host_mmio_locked(unsigned long phys, unsigned long size)
 {
-	u64 prot = host_mmu.pgt_ops->calc_pte_perm(true, true, true) |
-		   host_mmu.pgt_ops->calc_pte_memtype(true) |
+	u64 prot = host_mmu_pte_prot(true) |
 		   host_mmu.pgt_ops->pte_mk_pgstate(PKVM_PAGE_OWNED);
 	int ret;
 
--- a/arch/x86/kvm/pkvm/mmu.c
+++ b/arch/x86/kvm/pkvm/mmu.c
@@ -181,7 +181,7 @@ static int fix_hyp_mmu_page_refcnt(void)
 	 * Calculate the max address space, then walk the [0, size) address
 	 * range to fixup refcount of every page-table page.
 	 */
-	size = hyp_mmu.pgt_ops->level_to_size(hyp_mmu.cap.level + 1);
+	size = pkvm_pgtable_max_size(&hyp_mmu);
 #endif
 
 	return pkvm_pgtable_walk(&hyp_mmu, 0, size, &walker);
--- a/arch/x86/kvm/pkvm/mmu.c
+++ b/arch/x86/kvm/pkvm/mmu.c
@@ -338,11 +338,9 @@ static int fix_host_mmu_pgstate(void)
 
 static int host_mmu_map(unsigned long phys, unsigned long size, bool mmio)
 {
-	u64 prot = host_mmu.pgt_ops->calc_pte_perm(true, true, true) |
-		   host_mmu.pgt_ops->calc_pte_memtype(mmio);
-
 	/* The vaddr == phys for the host MMU */
-	return pkvm_pgtable_map(&host_mmu, phys, phys, size, prot);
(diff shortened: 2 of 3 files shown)


arch/x86/kvm/pkvm/mmu.c
```diff
@@ -190,13 +190,11 @@ static int fix_hyp_mmu_page_refcnt(void)
 static void set_host_mem_pgstate(unsigned long phys, unsigned long size,
 				 enum pkvm_page_state pgstate)
 {
+	unsigned long start = PAGE_ALIGN_DOWN(phys);
 	unsigned long end = PAGE_ALIGN(phys + size);
-	struct pkvm_page *page;
 
-	for (phys = PAGE_ALIGN_DOWN(phys); phys < end; phys += PAGE_SIZE) {
-		page = pkvm_phys_to_page(phys);
+	for_each_pkvm_page(page, start, end - start)
```
cxdong: nit: Suppose the `for_each_pkvm_page` macro will also be used at some other places in the code. Do the other places also need to calculate the size by `end - start`? If so, maybe it is better to just pass `end` to `for_each_pkvm_page` and let this macro calculate the size via something like `__ed - __st`?

arch/x86/kvm/pkvm/mmu.c
```diff
@@ -275,6 +275,12 @@ static int check_page_owner_and_state(struct pkvm_pgtable *pgt, unsigned long va
 	return pkvm_pgtable_walk(pgt, vaddr, size, &walker);
 }
 
+static u64 host_mmu_pte_prot(bool mmio)
+{
+	return host_mmu.pgt_ops->calc_pte_perm(true, true, true) |
```
cxdong, line 274: Just a note: the read, write, and exec are always true in this helper. This is sufficient for the current usages. In the future, if we want to map some read-only memory pages in the host mmu, this helper should be extended to take the detailed permissions as the inputs or create another helper for this purpose.

arch/x86/kvm/pkvm/mmu.c
maluka-dmytro: No, in those other places (in my WIP patches https://github.com/maluka-dmytro/pKVM-IA/commits/pvmmu-v6.18-wip) we already have the size, we don't need to calculate it. For example in `pkvm_host_share_guest()`: https://github.com/maluka-dmytro/pKVM-IA/commit/130524886f45efa8e39e3c06fd11a6f1f8535d43 and in other similar functions. In those cases the size is naturally passed as an argument to those API functions (and it is already page-aligned, unlike in host_share_hyp/hyp_share_host), so it is easier if they just pass it to `for_each_pkvm_page`, which is why I opted for passing the size rather than the end address.

> and let this macro calculate the size via something like `__ed - __st`?

If we just replace `__sz` with `__ed - __st`, it means we evaluate `__st` twice. (Although maybe we could just use `*__e = pkvm_phys_to_page(__ed)` instead.)

arch/x86/kvm/pkvm/mmu.c
maluka-dmytro: Yeah, good point. BTW I think we will actually have such cases with pvIOMMU.

Anyway, I guess it won't be a problem to add an extra argument to this helper for that, once we need it.

arch/x86/kvm/pkvm/mmu.c
cxdong: I see. I am fine with it as long as we are passing `__sz` for `for_each_pkvm_page` by some other places.

cxdong: > If we just replace __sz with __ed - __st, it means we evaluate __st twice. (Although maybe we could just use *__e = pkvm_phys_to_page(__ed) instead.)

This reminds me that, the current definition of `for_each_pkvm_page` requires the caller to pass in the page aligned start and size. If the caller forgets to do that, then there will be an issue. To avoid this potential issue, shall we align the start and size inside the macro, although this will introduce additional alignment if the start and size are already aligned?
```
#define for_each_pkvm_page(__p, __st, __sz)				\
	for (struct pkvm_page *__p = pkvm_phys_to_page(__st),		\
			      *__e = pkvm_phys_to_page(PAGE_ALIGN(__st + __sz));	\
	     __p < __e; __p++)
```

maluka-dmytro: Good point. Actually looks like if we do that, we can also simplify the code of the existing callers even more: they don't need to calculate `start` and `end`, they can just pass `phys` and `size` as is to `for_each_pkvm_page()` which will take care of alignment.

Although it means, again, that `__st` will be evaluated twice, i.e. `for_each_pkvm_page()` will need to be used with caution (the 2nd argument shouldn't have side effects). But we can just add a comment about that.

maluka-dmytro: Done.

arch/x86/kvm/pkvm/memory.h
```diff
@@ -44,6 +44,12 @@ extern u64 __pkvm_vmemmap;
 #define pkvm_page_to_virt(page)	__pkvm_va(pkvm_page_to_phys(page))
 #define pkvm_page_to_pool(page)	(((struct pkvm_page *)page)->pool)
 
+/* Caution: __st is evaluated twice. */
+#define for_each_pkvm_page(__p, __st, __sz)						\
+	for (struct pkvm_page *__p = pkvm_phys_to_page(PAGE_ALIGN_DOWN(__st)),		\
```
cxdong, line 49: nit: It seems we can remove `PAGE_ALIGN_DOWN` to simplify the first evaluation, as `pkvm_phys_to_page` will convert phys to pfn via `(phys) >> PAGE_SHIFT` which equals aligning down?

maluka-dmytro: Yes, but maybe it still makes sense to keep this `PAGE_ALIGN_DOWN`, just for code clarity?

cxdong: OK. That makes sense as well. I will merge this PR and squash these patches. Thanks!

PR #55, opened 2026-01-07
pkvm v6.18 fix MSR_CORE_PERF_GLOBAL_CTRL emulation and improve exception table sorting

cxdong:
The first two patches fixed the MSR_CORE_PERF_GLOBAL_CTRL emulation if it is not supported.

The last patch sorts the exception table before deprivileging. Previous the sorting is done during pkvm_init_finalize. As excpetion fixup doesn't rely on the initialization done by the pkvm_init_finalize, sort the exception table can be done earlier. Move it to the deprivilege phase so that the exception fixup mechanism can functional once the deprivileged is done.

### DIFF
--- a/arch/x86/kvm/vmx/pkvm_init.c
+++ b/arch/x86/kvm/vmx/pkvm_init.c
@@ -3,6 +3,7 @@
 
 #include <linux/kernel.h>
 #include <linux/module.h>
+#include <linux/extable.h>
 #include <asm/pkvm_image.h>
 #include "pkvm_constants.h"
 #include "vmx.h"
@@ -1025,6 +1026,16 @@ static __init int pkvm_host_deprivilege_cpus(struct pkvm_hyp *pkvm)
 	int cpu, ret = 0;
 
 	pkvm_sym(pkvm_vmx_register_excp_handlers)();
+
+	/*
+	 * The pKVM hypervisor's IDT will be programmed into VMCS before
+	 * deprivileging the CPU. Once deprivileging is done and the CPU
+	 * enters to the root mode, the pKVM's exception handlers should be
+	 * functional. So before that, sort pKVM's exception table to make
+	 * sure the exception fixup working as expected.
+	 */
+	if (&pkvm_sym(__stop___ex_table) > &pkvm_sym(__start___ex_table))
+		sort_extable(pkvm_sym(__start___ex_table), pkvm_sym(__stop___ex_table));
 
 	for_each_possible_cpu(cpu) {
 		ret = smp_call_function_single(cpu, pkvm_host_deprivilege_cpu, &p, 1);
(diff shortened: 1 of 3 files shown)


arch/x86/kvm/vmx/pkvm_init.c
```diff
@@ -714,7 +714,11 @@ static __init void init_guest_state_area_from_native(struct vcpu_vmx *vmx)
 
 	if (!rdmsrq_safe(MSR_CORE_PERF_GLOBAL_CTRL, &msrq)) {
 		struct kvm_pmu *pmu = vcpu_to_pmu(&vmx->vcpu);
+		struct x86_pmu_capability pmu_cap;
 
+		perf_get_x86_pmu_capability(&pmu_cap);
```
maluka-dmytro: The PR looks good to me, except that I see a small issue: on a hybrid CPU (e.g. on my ARL), when booting, this is causing a call trace printed during bootup, due to this WARN_ON_ONCE:

```
void perf_get_x86_pmu_capability(struct x86_pmu_capability *cap)
{
	/* This API doesn't currently support enumerating hybrid PMUs. */
	if (WARN_ON_ONCE(cpu_feature_enabled(X86_FEATURE_HYBRID_CPU)) ||
	    !x86_pmu_initialized()) {
		memset(cap, 0, sizeof(*cap));
		return;
	}
```

So maybe we can e.g. just do `pmu->version = x86_pmu.version` instead?

cxdong: I tried to use x86_pmu.version directly, but seems x86_pmu is decleared in arch/x86/events/perf_event.h, which is not for the KVM subsystem, so cannot use it. Maybe we need to get the version from CPUID directly.

maluka-dmytro: Actually it is causing not just printing a call trace, it is causing not returning the pmu caps (so `pmu_cap.version` is 0).

cxdong: Yeah, I didn't realize this doesn't work for the hybrid CPU case. Directly getting from the CPUID 10 should work. Will update with this.

maluka-dmytro: BTW am I correct that we don't need to precisely know the PMU version, we just need to know if MSR_CORE_PERF_GLOBAL_CTRL is available? And here, since rdmsrq_safe succeeded, we already know it is available, so we could just remember that in some flag, to use it in `handle_write_msr()`? Or perhaps we could just directly probe it via `rdmsrq_safe` again in `handle_write_msr()` every time (although that may be slower)?

maluka-dmytro: Anyway, checked that the latest version seems to work fine on ARL (and the obtained `pmu->version` is 5).

cxdong: > BTW am I correct that we don't need to precisely know the PMU version, we just need to know if MSR_CORE_PERF_GLOBAL_CTRL is available?

Yes it is correct.

>And here, since rdmsrq_safe succeeded, we already know it is available, so we could just remember that in some flag, to use it in handle_write_msr()? Or perhaps we could just directly probe it via rdmsrq_safe again in handle_write_msr() every time (although that may be slower)?

These should work too. But any reason prefer these rather than setting `pmu->version`? It seems that setting `pmu->version` can reuse the existing mechanism comparing with the way of adding new flag, and is faster than probing it via `rdmsr_safe()` in `handle_write_msr()` every time.

maluka-dmytro: Yeah, both ways seems simple enough, so I'm ok with keeping the current way.

maluka-dmytro (approved):

PR #56, opened 2026-01-13
SQUASHME: pKVM: x86: Inject #GP for the host lapic emulation failure

cxdong:
Inject #GP to the host if the LAPIC emulation is failed rather than BUG_ON in the pKVM hypervisor.

Fixes: bc395be4431f ("pKVM: VMX: Intercept writing to some X2APIC MSRs")


maluka-dmytro:
> rather than BUG_ON in the pKVM hypervisor.

You probably mean not a BUG_ON but `wrmsr_safe(msr, low, high)` at line 198?

Anyway, the fix looks good to me.

cxdong:
It will triger BUG_ON first so will not go to `wrmsr_safe` at line  198.

maluka-dmytro:
Ah, indeed, it will trigger BUG_ON since APIC MSRs are within the bitmap range. Anyway, nevermind.

maluka-dmytro (approved):

PR #48, opened 2025-12-09
Pkvm v6.18 pvvmcs part2

cxdong:
pvvmcs part2 to implement the PV interfaces of accessing the vcpu state

PR #57, opened 2026-01-14
pKVM: x86: Simplify extable support

maluka-dmytro:
We don't sort pKVM's extable on the hypervisor side anymore (we do that on the host side before deprivileging). So we don't need to compile `lib/sort.c` into the hypervisor code, and thus also can avoid making pKVM-specific changes to `lib/sort.c`.

Also this means that only a small part of `lib/extable.c` is used by the hypervisor, so it's easier to copy that small part, instead of pulling in the entire `lib/extable.c`.

### DIFF
--- a/arch/x86/kvm/pkvm/Makefile
+++ b/arch/x86/kvm/pkvm/Makefile
@@ -29,8 +29,7 @@ quiet_cmd_patch_retpoline = PATCH_RETPOLINE $@
       cmd_patch_retpoline = $(BASH) $(src)/compiling_cmds.sh patch_retpoline $< $@
 
 kernel-lib			:= ../../../../lib
-pkvm-hyp-y			+= $(kernel-lib)/sort.o $(kernel-lib)/bsearch.o \
-				   $(kernel-lib)/extable.o $(kernel-lib)/find_bit.o \
+pkvm-hyp-y			+= $(kernel-lib)/bsearch.o $(kernel-lib)/find_bit.o \
 				   $(kernel-lib)/hweight.o $(kernel-lib)/string.o \
 				   $(kernel-lib)/ctype.o
 pkvm-hyp-$(CONFIG_LIST_HARDENED) += $(kernel-lib)/list_debug.o
--- a/lib/sort.c
+++ b/lib/sort.c
@@ -264,10 +264,6 @@ static void __sort_r(void *base, size_t num, size_t size,
 			do_swap(base + b, base + c, size, swap_func, priv);
 		}
 
-#ifndef __PKVM_HYP__
-		if (may_schedule)
-			cond_resched();
-#else
 		/*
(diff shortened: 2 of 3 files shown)


arch/x86/kvm/pkvm/Makefile
```diff
@@ -29,8 +29,7 @@ quiet_cmd_patch_retpoline = PATCH_RETPOLINE $@
       cmd_patch_retpoline = $(BASH) $(src)/compiling_cmds.sh patch_retpoline $< $@
 
 kernel-lib			:= ../../../../lib
-pkvm-hyp-y			+= $(kernel-lib)/sort.o $(kernel-lib)/bsearch.o \
-				   $(kernel-lib)/extable.o $(kernel-lib)/find_bit.o \
+pkvm-hyp-y			+= $(kernel-lib)/bsearch.o $(kernel-lib)/find_bit.o \
```
cxdong, line 32: Actually commit https://github.com/intel-staging/pKVM-IA/pull/48/commits/65498d0ebf8c684ba5b70f60547354330bbee669 in pvvmcs part 2 is also using the sort lib. The reason is that, it import the function `setup_xstate_cache()` from arch/x86/kernel/fpu/xstate.c to setup the `xstate_offsets`, `xstate_sizes`, `xstate_flags` and `xfeature_uncompact_order` which will be used by the pKVM. And this function will call `sort()`. If we remove the sort lib, maybe we also need to change this commit.

Per the discussion in the chat room, seems it is better for us not to change the code in include/ and lib/ folders? If so, we can take this change, and need to have some alternatives for commit https://github.com/intel-staging/pKVM-IA/pull/48/commits/65498d0ebf8c684ba5b70f60547354330bbee66

lib/sort.c
```diff
@@ -264,18 +264,13 @@ static void __sort_r(void *base, size_t num, size_t size,
 			do_swap(base + b, base + c, size, swap_func, priv);
 		}
 
-#ifndef __PKVM_HYP__
-		if (may_schedule)
-			cond_resched();
-#else
 		/*
```
cxdong, line 267: We should keep the code in `#ifndef __PKVM_HYP__` but remove the code below?

arch/x86/kvm/pkvm/Makefile
cxdong: How about we just copy this lib/sort.c to arch/x86/kvm/pkvm/lib/ as pKVM's own file? It most likely we don't need to update this file in the future so no maintianing burdon to keep this file in sync with the upstream lib/sort.c during the kernel upgrading.

lib/sort.c
maluka-dmytro: Ah yes, sorry...

cxdong:
Will combine this PR with https://github.com/intel-staging/pKVM-IA/pull/48, so close this one.

PR #59, opened 2026-01-16
Pkvm v6.18 cleanup non x86

cxdong:


cxdong:
Hi @jaszczyk-grzegorz @maluka-dmytro,
1. The first 2 commits are fixing some compiling issue. As we decide to squash this PR, I feel we can also squash these two patches.
2. Two patches from Greg at this branch https://github.com/maluka-dmytro/pKVM-IA/commits/pkvm-v6.18-header-changes-reduction/ are merged as https://github.com/intel-staging/pKVM-IA/pull/59/commits/1458f9ad4bd889a1a173f76ffeadc643b9cab31b
3. One patch from Dmia from https://github.com/maluka-dmytro/pKVM-IA/commits/pkvm-v6.18-header-changes-reduction/ is commit https://github.com/intel-staging/pKVM-IA/pull/59/commits/ca10037478ea697eb600d7ace6638651865c76aa
4. The other patches from Dima are not included in this PR as they are for pvVMCS part 2. I will consolidate them with pvVMCS part2 after this PR is squashed.
5. One major change is introduced a kern_stub.c, to implement some stub symbols to avoid changing generic linux headers.
6. As commit https://github.com/intel-staging/pKVM-IA/pull/59/commits/b4ed21f5ac94744c81654bedfe82d4a6a119f95f has removed the changes in sort.c and it is imported, so it seems we can just keep importing extable.c rather than removing it by the commit https://github.com/intel-staging/pKVM-IA/commit/ec421bc8354167abbf243ede6b942beb394e743b

jaszczyk-grzegorz:
In general LGTM (but will wait on @maluka-dmytro feedback) and thanks for making those changes it looks more compact now and it is nice to see only:
```
 include/linux/export.h                          |    2 +-
 tools/arch/x86/include/uapi/asm/kvm.h           |    6 +
 tools/objtool/arch/x86/decode.c                 |    5 +-
 tools/objtool/check.c                           |    1 +
 tools/objtool/special.c                         |    6 +
```
modification outside arch/x86.

One thing is that maybe it will be a bit cleaner to apply also: https://github.com/jaszczyk-grzegorz/pKVM-IA/commit/abad4cfd5d319e761e91540ac15335f9c469b317 instead of squashing it especially that this is 1:1 copy with the copyrights and so on - it clearly shows the original file and the diff we introduced. But if you and @maluka-dmytro think having one patch is better I am fine with that.
BTW you are planning to reorder this commit before, rather than squash with "pKVM: x86: Import the kernel find_bit and hweight libraries", and import hweight lib separately, right?

arch/x86/include/asm/msr.h
```diff
@@ -43,7 +43,12 @@ struct saved_msrs {
 #include <asm/atomic.h>
 #include <linux/tracepoint-defs.h>
 
-#if defined(CONFIG_TRACEPOINTS) && !defined(__PKVM_HYP__)
+#if defined(__PKVM_HYP__) && defined(CONFIG_TRACEPOINTS)
```
maluka-dmytro, line 46: Just a note: this is still a non-KVM header, so ideally it would be good to avoid changing it as well. But if that's not easy, we can leave it for now.

arch/x86/kernel/vmlinux.lds.S
```diff
@@ -159,15 +159,21 @@ PHDRS {
 	pkvm_sym(rodata_end) = .;						\
 	}
 
+/*
+ * The PKVM_PERCPU requires to be put right after the PERCPU_SECION. It will
```
maluka-dmytro: nit: s/SECION/SECTION/

arch/x86/include/asm/kvm_pkvm.h
```diff
@@ -373,6 +374,34 @@ static inline size_t pkvm_guest_initial_fpstate_size(struct kvm *kvm)
...
+#define WARN(condition, format...) ({					\
+	int __ret_warn_on = !!(condition);				\
+	no_printk(format);						\
+	unlikely(__ret_warn_on);					\
+})
+
+#define WARN_ON_ONCE(condition) WARN_ON(condition)
+#define WARN_ONCE(condition, format...) WARN(condition, format)
+
+#define _BUG_FLAGS(ins, flags, extra)  asm volatile(ins)
```
jaszczyk-grzegorz, line 400: checkpatch throws error on that, it should be surrounded with parentheses:
```
ERROR: Macros with complex values should be enclosed in parentheses                                                                                                                                                                                                                                                            
#400: FILE: arch/x86/include/asm/kvm_pkvm.h:400:                                                                                                                                                                                                                                                                               
+#define _BUG_FLAGS(ins, flags, extra)  asm volatile(ins)                                                                                                                                                                                                                                                                      
                                                                                                                                                                                                                                                                                                                               
BUT SEE:                                                                                                                                                                                                                                                                                                                       
                                                                                                                                                                                                                                                                                                                               
   do {} while (0) advice is over-stated in a few situations:                                                                                                                                                                                                                                                                  
                                                                                                                                                                                                                                                                                                                               
   The more obvious case is macros, like MODULE_PARM_DESC, invoked at                                                                                                                                                                                                                                                          
   file-scope, where C disallows code (it must be in functions).  See                                                                                                                                                                                                                                                          
   $exceptions if you have one to add by name.                                                                                                                                                                                                                                                                                 
                                                                                                                                                                                                                                                                                                                               
   More troublesome is declarative macros used at top of new scope,                                                                                                                                                                                                                                                            
   like DECLARE_PER_CPU.  These might just compile with a do-while-0
   wrapper, but would be incorrect.  Most of these are handled by
   detecting struct,union,etc declaration primitives in $exceptions.

   Theres also macros called inside an if (block), which "return" an
   expression.  These cannot do-while, and need a ({}) wrapper.

   Enjoy this qualification while we work to improve our heuristics.
```

maluka-dmytro:
> One thing is that maybe it will be a bit cleaner to apply also: [jaszczyk-grzegorz@abad4cf](https://github.com/jaszczyk-grzegorz/pKVM-IA/commit/abad4cfd5d319e761e91540ac15335f9c469b317) instead of squashing it especially that this is 1:1 copy with the copyrights and so on - it clearly shows the original file and the diff we introduced. But if you and @maluka-dmytro think having one patch is better I am fine with that.

I'm fine either way.

arch/x86/include/asm/kvm_pkvm.h
maluka-dmytro: This definition of _BUG_FLAGS is actually copied from `arch/x86/include/asm/bug.h`

So should be ok to keep it as is?

maluka-dmytro:
```
dmaluka@dmaluka:~/src/pKVM-IA$ grep -r 'define.*asm volatile' arch/x86/include/
arch/x86/include/asm/barrier.h:#define mb() asm volatile(ALTERNATIVE("lock addl $0,-4(%%esp)", "mfence", \
arch/x86/include/asm/barrier.h:#define rmb() asm volatile(ALTERNATIVE("lock addl $0,-4(%%esp)", "lfence", \
arch/x86/include/asm/barrier.h:#define wmb() asm volatile(ALTERNATIVE("lock addl $0,-4(%%esp)", "sfence", \
arch/x86/include/asm/barrier.h:#define __mb()	asm volatile("mfence":::"memory")
arch/x86/include/asm/barrier.h:#define __rmb()	asm volatile("lfence":::"memory")
arch/x86/include/asm/barrier.h:#define __wmb()	asm volatile("sfence" ::: "memory")
arch/x86/include/asm/barrier.h:#define __smp_mb()	asm volatile("lock addl $0,-4(%%" _ASM_SP ")" ::: "memory", "cc")
arch/x86/include/asm/kvm_pkvm.h:#define _BUG_FLAGS(ins, flags, extra)  asm volatile(ins)
arch/x86/include/asm/bug.h:#define _BUG_FLAGS(ins, flags, extra)  asm volatile(ins)
arch/x86/include/asm/desc.h:#define load_tr(tr)				asm volatile("ltr %0"::"m" (tr))
arch/x86/include/asm/desc.h:#define load_ldt(ldt)				asm volatile("lldt %0"::"m" (ldt))
arch/x86/include/asm/sev.h:#define	VMGEXIT()			{ asm volatile("rep; vmmcall\n\r"); }
arch/x86/include/asm/special_insns.h:#define nop() asm volatile ("nop")
```

arch/x86/include/asm/pkvm_trace.h
```diff
@@ -7,9 +7,11 @@
 #include <asm/vmx.h>
 
 /*
- * TODO: the implementation of vmexit tracing is VMX-specific.
- * Make it generic.
+ * TODO: The implementation of vmexit tracing is VMX-specific. Make it generic.
+ * If there is new vmx vmexit reason in the future which is larger than the
```
maluka-dmytro: nit: this sentence is not quite related to the above `TODO:`, so maybe just add another `TODO:` before this sentence? So that it's clear that these are two separate TODOs.

arch/x86/include/asm/kvm_pkvm.h
jaszczyk-grzegorz: Ok, I see - yeah I think it should be ok then, just will need to bypass checkpatch check later on but it shouldn't be an issue.

cxdong:
> In general LGTM (but will wait on @maluka-dmytro feedback) and thanks for making those changes it looks more compact now and it is nice to see only:
> 
> ```
>  include/linux/export.h                          |    2 +-
>  tools/arch/x86/include/uapi/asm/kvm.h           |    6 +
>  tools/objtool/arch/x86/decode.c                 |    5 +-
>  tools/objtool/check.c                           |    1 +
>  tools/objtool/special.c                         |    6 +
> ```
> 
> modification outside arch/x86.
> 
> One thing is that maybe it will be a bit cleaner to apply also: [jaszczyk-grzegorz@abad4cf](https://github.com/jaszczyk-grzegorz/pKVM-IA/commit/abad4cfd5d319e761e91540ac15335f9c469b317) instead of squashing it especially that this is 1:1 copy with the copyrights and so on - it clearly shows the original file and the diff we introduced. 

From this point of view, it makes more sense to keep this 1:1 copy patch with the copyrights then apply the next patch on the top. @maluka-dmytro is fine with either way, so I will add this 1:1 copy patch back.

>But if you and @maluka-dmytro think having one patch is better I am fine with that. BTW you are planning to reorder this commit before, rather than squash with "pKVM: x86: Import the kernel find_bit and hweight libraries", and import hweight lib separately, right?

Yes, will import them separtely. Probably import hweight lib first as it is needed by find_bit.

arch/x86/include/asm/pkvm_trace.h
cxdong: Sure, will separete TODOs.

arch/x86/kernel/vmlinux.lds.S
cxdong: Will fix this.

arch/x86/include/asm/msr.h
cxdong: Yeah, didn't find a better way to redefine this inside the KVM's header file. Probably can leave it for now if GKI maintainers are fine with it.

maluka-dmytro (approved):

maluka-dmytro:
LGTM and works well in my tests. IMHO we can go ahead and squash it.

maluka-dmytro:
For the record: the issue fixed by https://github.com/intel-staging/pKVM-IA/pull/59/commits/ee144a5b376ebb4cb3c7466ad7902778f01ed75e ("SQUASHME: pKVM: x86: Enable objtool when compiling pKVM objects") was reproduced with both CONFIG_X86_KERNEL_IBT and CONFIG_LTO_CLANG disabled. In such case, the issue was: there is "Unpatched return thunk in use." warning caused by vmx_pkvm_init() during boot, and then, if in non-debug pKVM mode, the system hangs after deprivileging.

jaszczyk-grzegorz (approved):
LGTM

cxdong:
> For the record: the issue fixed by [ee144a5](https://github.com/intel-staging/pKVM-IA/commit/ee144a5b376ebb4cb3c7466ad7902778f01ed75e) ("SQUASHME: pKVM: x86: Enable objtool when compiling pKVM objects") was reproduced with both CONFIG_X86_KERNEL_IBT and CONFIG_LTO_CLANG disabled. In such case, the issue was: there is "Unpatched return thunk in use." warning caused by vmx_pkvm_init() during boot, and then, if in non-debug pKVM mode, the system hangs after deprivileging.

Yes, excatly.

PR #63, opened 2026-01-22
pKVM: x86: Support pKVM hypercalls with pkvm_hc_data as input and output

cxdong:
Add macro pkvm_hypercall_inout to support sending pKVM hypercalls with using union pkvm_hc_data as both inputs and outputs. Such hypercalls should be defined by PKVM_HC_INOUT in the pkvm_hypercalls.h. And its corresponding structure in union pkvm_hc_data should be defined as something like:
    union pkvm_hc_data {
	    ......
	    union {
		    struct {
			    ......
		    } in;
		    struct {
			    ......
		    } out;
	    } hypercall_name;
    };
So that the size of the input and output can be different.

### DIFF
--- a/arch/x86/include/asm/kvm_pkvm.h
+++ b/arch/x86/include/asm/kvm_pkvm.h
@@ -132,17 +132,23 @@ static_assert(sizeof(union pkvm_hc_data) == PKVM_HC_DATA_MAX_NUM * sizeof(u64));
 
 #define PKVM_HC_DATA_NUM(f)		\
 	(ALIGN(sizeof(((union pkvm_hc_data *)0)->f), sizeof(u64)) / sizeof(u64))
+#define PKVM_HC_DATA_OUT_NUM(f)		\
+	(ALIGN(sizeof(((union pkvm_hc_data *)0)->f.out), sizeof(u64)) / sizeof(u64))
+#define PKVM_HC_DATA_IN_NUM(f)		\
+	(ALIGN(sizeof(((union pkvm_hc_data *)0)->f.in), sizeof(u64)) / sizeof(u64))
 
 #define PKVM_HC_OUTPUT_NUM(f)	f##_output_num
 #define PKVM_HC_INPUT_NUM(f)	f##_input_num
 
 enum {
-	#define PKVM_HC(f)	PKVM_HC_OUTPUT_NUM(f) = 0,
-	#define PKVM_HC_OUT(f)	PKVM_HC_OUTPUT_NUM(f) = PKVM_HC_DATA_NUM(f),
+	#define PKVM_HC(f)		PKVM_HC_OUTPUT_NUM(f) = 0,
+	#define PKVM_HC_OUT(f)		PKVM_HC_OUTPUT_NUM(f) = PKVM_HC_DATA_NUM(f),
+	#define PKVM_HC_INOUT(f)	PKVM_HC_OUTPUT_NUM(f) = PKVM_HC_DATA_OUT_NUM(f),
 	#include <asm/pkvm_hypercalls.h>
 
-	#define PKVM_HC(f)	PKVM_HC_INPUT_NUM(f) = 0,
-	#define PKVM_HC_IN(f)	PKVM_HC_INPUT_NUM(f) = PKVM_HC_DATA_NUM(f),
+	#define PKVM_HC(f)		PKVM_HC_INPUT_NUM(f) = 0,
+	#define PKVM_HC_IN(f)		PKVM_HC_INPUT_NUM(f) = PKVM_HC_DATA_NUM(f),
+	#define PKVM_HC_INOUT(f)	PKVM_HC_INPUT_NUM(f) = PKVM_HC_DATA_IN_NUM(f),
 	#include <asm/pkvm_hypercalls.h>
 };
 
@@ -201,7 +207,7 @@ static inline int pkvm_hc_input_num(enum pkvm_hc hc)
 	__pkvm_hypercall(f, NULL, 0, ##__VA_ARGS__);					\
 })
 
-#define pkvm_hypercall_out(f, o, ...)							\
+#define __pkvm_hypercall_inout(f, o, ...)						\
 	__builtin_choose_expr(PKVM_HC_OUTPUT_NUM(f) == 0,				\
 			      __pkvm_hypercall(f, NULL, 0, ##__VA_ARGS__),		\
 	__builtin_choose_expr(PKVM_HC_OUTPUT_NUM(f) == 1,				\
@@ -214,22 +220,28 @@ static inline int pkvm_hc_input_num(enum pkvm_hc hc)
 			      __pkvm_hypercall(f, o, 4, ##__VA_ARGS__),			\
 	PKVM_HC_UNREACHABLE(f))))))
 
-#define pkvm_hypercall_in(f, i)								\
+#define pkvm_hypercall_inout(f, i, o)							\
 	__builtin_choose_expr(PKVM_HC_INPUT_NUM(f) == 1,				\
-			      __pkvm_hypercall(f, NULL, 0, (i)->raw.data[0]),		\
+			      __pkvm_hypercall_inout(f, o, (i)->raw.data[0]),		\
 	__builtin_choose_expr(PKVM_HC_INPUT_NUM(f) == 2,				\
-			      __pkvm_hypercall(f, NULL, 0, (i)->raw.data[0],		\
-					       (i)->raw.data[1]),			\
+			      __pkvm_hypercall_inout(f, o, (i)->raw.data[0],		\
+						     (i)->raw.data[1]),			\
 	__builtin_choose_expr(PKVM_HC_INPUT_NUM(f) == 3,				\
-			      __pkvm_hypercall(f, NULL, 0, (i)->raw.data[0],		\
-					       (i)->raw.data[1],			\
-					       (i)->raw.data[2]),			\
+			      __pkvm_hypercall_inout(f, o, (i)->raw.data[0],		\
+						     (i)->raw.data[1],			\
+						     (i)->raw.data[2]),			\
 	__builtin_choose_expr(PKVM_HC_INPUT_NUM(f) == 4,				\
-			      __pkvm_hypercall(f, NULL, 0, (i)->raw.data[0],		\
-					       (i)->raw.data[1],			\
-					       (i)->raw.data[2],			\
-					       (i)->raw.data[3]),			\
+			      __pkvm_hypercall_inout(f, o, (i)->raw.data[0],		\
+						     (i)->raw.data[1],			\
+						     (i)->raw.data[2],			\
+						     (i)->raw.data[3]),			\
 	PKVM_HC_UNREACHABLE(f)))))
+
+#define pkvm_hypercall_out(f, o, ...)							\
+	__pkvm_hypercall_inout(f, o, ##__VA_ARGS__)
+
+#define pkvm_hypercall_in(f, i)								\
+	pkvm_hypercall_inout(f, i, (union pkvm_hc_data *)NULL)
 
 static inline unsigned long pkvm_hc(struct kvm_vcpu *vcpu)
 {
--- a/arch/x86/include/asm/kvm_pkvm.h
+++ b/arch/x86/include/asm/kvm_pkvm.h
@@ -61,7 +61,7 @@ enum pkvm_hc {
 	MAX_PKVM_HYPERCALLS,
 };
 
-#define PKVM_HC_DATA_MAX_NUM		4
+#define PKVM_HC_DATA_MAX_NUM		10
 
 union pkvm_hc_data {
 	struct {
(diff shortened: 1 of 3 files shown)


arch/x86/include/asm/kvm_pkvm.h
```diff
@@ -132,17 +132,23 @@ static_assert(sizeof(union pkvm_hc_data) == PKVM_HC_DATA_MAX_NUM * sizeof(u64));
 
 #define PKVM_HC_DATA_NUM(f)		\
 	(ALIGN(sizeof(((union pkvm_hc_data *)0)->f), sizeof(u64)) / sizeof(u64))
+#define PKVM_HC_DATA_OUT(f)		\
+	(ALIGN(sizeof(((union pkvm_hc_data *)0)->f.out), sizeof(u64)) / sizeof(u64))
+#define PKVM_HC_DATA_IN(f)		\
+	(ALIGN(sizeof(((union pkvm_hc_data *)0)->f.in), sizeof(u64)) / sizeof(u64))
```
maluka-dmytro, line 138: Probably PKVM_HC_DATA_OUT_NUM and PKVM_HC_DATA_IN_NUM?

arch/x86/include/asm/kvm_pkvm.h
```diff
@@ -214,23 +220,29 @@ static inline int pkvm_hc_input_num(enum pkvm_hc hc)
...
-			      __pkvm_hypercall(f, NULL, 0, (i)->raw.data[0],		\
+			      pkvm_hypercall_out(f, o, (i)->raw.data[0],		\
 					       (i)->raw.data[1],			\
 					       (i)->raw.data[2],			\
 					       (i)->raw.data[3]),			\
 	PKVM_HC_UNREACHABLE(f)))))
 
+#define pkvm_hypercall_in(f, i)								\
+({											\
+	union pkvm_hc_data __out;							\
```
vineethrp: do we need this here? can we pass NULL instead?

arch/x86/include/asm/kvm_pkvm.h
```diff
@@ -214,23 +220,29 @@ static inline int pkvm_hc_input_num(enum pkvm_hc hc)
 			      __pkvm_hypercall(f, o, 4, ##__VA_ARGS__),			\
 	PKVM_HC_UNREACHABLE(f))))))
 
-#define pkvm_hypercall_in(f, i)								\
+#define pkvm_hypercall_inout(f, i, o)							\
 	__builtin_choose_expr(PKVM_HC_INPUT_NUM(f) == 1,				\
-			      __pkvm_hypercall(f, NULL, 0, (i)->raw.data[0]),		\
+			      pkvm_hypercall_out(f, o, (i)->raw.data[0]),		\
```
vineethrp: nit: I think its more readable to have __pkvm_hypercall itself here and less changes. WDYT?

arch/x86/include/asm/kvm_pkvm.h
cxdong: Yeah, I can update that.

arch/x86/include/asm/kvm_pkvm.h
cxdong: There will be compiling issue if we pass NULL here. It seems the macro `pkvm_hypercall_inout` may be expanded with NULL?

arch/x86/include/asm/kvm_pkvm.h
cxdong: The issue is that, `__pkvm_hypercall` expects to have a const output number which can be got in the pre-processing stage. If we use `__pkvm_hypercall` here, then we need to replace `0` with `PKVM_HC_OUTPUT_NUM(f)` (unless we assume the number of inputs always equals the number of outputs?) which cannot be got  the pre-processing stage, thus causing the compiling issue.

arch/x86/include/asm/kvm_pkvm.h
maluka-dmytro: I haven't carefully looked into how this works yet, but I see that the `__out` value is not initialized?

I _guess_ this value is not actually used so that's ok, but still seems not very clean?

...And just a thought: maybe it would be simpler or cleaner to just keep the existing implementations of `pkvm_hypercall_in()` and `pkvm_hypercall_out()` as is, and just manually implement `pkvm_hypercall_inout()` similarly to those existing implementations, without reusing or modifying them?

vineethrp: Sorry I meant NULL (or 0) with a cast (union pkvm_hc_data *). I tried it and compiled for me:

```
diff --git a/arch/x86/include/asm/kvm_pkvm.h b/arch/x86/include/asm/kvm_pkvm.h
index 14b47c7161f25..59d197f72aafb 100644
--- a/arch/x86/include/asm/kvm_pkvm.h
+++ b/arch/x86/include/asm/kvm_pkvm.h
@@ -248,8 +248,7 @@ static inline int pkvm_hc_input_num(enum pkvm_hc hc)
 
 #define pkvm_hypercall_in(f, i)                                                                \
 ({                                                                                     \
-       union pkvm_hc_data __out;                                                       \
-       pkvm_hypercall_inout(f, i, &__out);                                             \
+       pkvm_hypercall_inout(f, i, (union pkvm_hc_data *)NULL);                                         \
 })
 
 static inline unsigned long pkvm_hc(struct kvm_vcpu *vcpu)
```

Since HC_IN hypercalls are more used than HC_INOUT, I thought we could optimize it like this. But I do not have a strong preference

arch/x86/include/asm/kvm_pkvm.h
vineethrp: Yeah true we would have to use PKVM_HC_OUTPUT_NUM(f) and incurs changing this anyways. I am fine wit this.

arch/x86/include/asm/kvm_pkvm.h
maluka-dmytro: All this still looks a bit non-intuitive to me: `pkvm_hypercall_in()` is implemented via `pkvm_hypercall_inout()`, which is implemented via `pkvm_hypercall_out()` (so `pkvm_hypercall_in()` is implemented via `pkvm_hypercall_out()`)? :)

cxdong: Probably the intutive way is to keep the existing `pkvm_hypercall_in` and `pkvm_hypercall_out`, and define a new macr `pkvm_hypercall_inout`, which is implemented by `pkvm_hypercall_out`, as `pkvm_hypercall_out` has both inputs and outputs?
```
+#define pkvm_hypercall_inout(f, i, o)                                                  \
+       __builtin_choose_expr(PKVM_HC_INPUT_NUM(f) == 1,                                \
+                             pkvm_hypercall_out(f, o, (i)->raw.data[0]),               \
+       __builtin_choose_expr(PKVM_HC_INPUT_NUM(f) == 2,                                \
+                             pkvm_hypercall_out(f, o, (i)->raw.data[0],                \
+                                              (i)->raw.data[1]),                       \
+       __builtin_choose_expr(PKVM_HC_INPUT_NUM(f) == 3,                                \
+                             pkvm_hypercall_out(f, o, (i)->raw.data[0],                \
+                                              (i)->raw.data[1],                        \
+                                              (i)->raw.data[2]),                       \
+       __builtin_choose_expr(PKVM_HC_INPUT_NUM(f) == 4,                                \
+                             pkvm_hypercall_out(f, o, (i)->raw.data[0],                \
+                                              (i)->raw.data[1],                        \
+                                              (i)->raw.data[2],                        \
+                                              (i)->raw.data[3]),                       \
+       PKVM_HC_UNREACHABLE(f)))))
+

```

vineethrp: Yeah this seems more intuitive and readable IMHO

maluka-dmytro (approved):

maluka-dmytro:
Now it looks nice and simple to me.

I guess we don't need to merge it now, we can merge it later together with pvIOMMU patches which will actually use it?

cxdong:
> Now it looks nice and simple to me.
> 
> I guess we don't need to merge it now, we can merge it later together with pvIOMMU patches which will actually use it?

Yes, I also plan to keep it open until it is needed.

vineethrp:
Looks good to me as well!

> I guess we don't need to merge it now, we can merge it later together with pvIOMMU patches which will actually use it?

I am fine with this. I shall be cherry-picking this on the pkvm initialization series today

arch/x86/include/asm/kvm_pkvm.h
```diff
@@ -172,22 +172,105 @@ static inline int pkvm_hc_input_num(enum pkvm_hc hc)
...
+#define PKVM_HC_IN_3(...)		PKVM_HC_IN_2(__VA_ARGS__), "d"(IN_ARG_3(__VA_ARGS__))
+#define PKVM_HC_IN_4(...)		PKVM_HC_IN_3(__VA_ARGS__), "S"(IN_ARG_4(__VA_ARGS__))
+#define PKVM_HC_IN_5(...)		PKVM_HC_IN_4(__VA_ARGS__), "D"(IN_ARG_5(__VA_ARGS__))
+#define PKVM_HC_IN_6(...)		PKVM_HC_IN_5(__VA_ARGS__), "r" (r8_in)
+#define PKVM_HC_IN_7(...)		PKVM_HC_IN_6(__VA_ARGS__), "r" (r9_in)
+#define PKVM_HC_IN_8(...)		PKVM_HC_IN_7(__VA_ARGS__), "r" (r10_in)
+#define PKVM_HC_IN_9(...)		PKVM_HC_IN_8(__VA_ARGS__), "r" (r11_in)
+#define PKVM_HC_IN_10(...)		PKVM_HC_IN_9(__VA_ARGS__), "r" (r12_in)
+
+#define PKVM_HC_DECLEAR_OUT_0
```
maluka-dmytro: s/DECLEAR/DECLARE/ ?

arch/x86/include/asm/kvm_pkvm.h
```diff
@@ -172,22 +172,105 @@ static inline int pkvm_hc_input_num(enum pkvm_hc hc)
...
+#define IN_ARG_3(a1, a2, a3, ...)					(unsigned long)a3
+#define IN_ARG_4(a1, a2, a3, a4, ...)					(unsigned long)a4
+#define IN_ARG_5(a1, a2, a3, a4, a5, ...)				(unsigned long)a5
+#define IN_ARG_6(a1, a2, a3, a4, a5, a6, ...)				(unsigned long)a6
+#define IN_ARG_7(a1, a2, a3, a4, a5, a6, a7, ...)			(unsigned long)a7
+#define IN_ARG_8(a1, a2, a3, a4, a5, a6, a7, a8, ...)			(unsigned long)a8
+#define IN_ARG_9(a1, a2, a3, a4, a5, a6, a7, a8, a9, ...)		(unsigned long)a9
+#define IN_ARG_10(a1, a2, a3, a4, a5, a6, a7, a8, a9, a10, ...)		(unsigned long)a10
+
+#define PKVM_HC_DECLEAR_IN_0(...)
```
vineethrp: s/PKVM_HC_DECLEAR_{IN,OUT}_*/PKVM_HC_DECLARE_{IN,OUT}_*/?

arch/x86/include/asm/kvm_pkvm.h
```diff
@@ -172,22 +172,105 @@ static inline int pkvm_hc_input_num(enum pkvm_hc hc)
...
+#define PKVM_HC_DECLEAR_IN_1(...)	PKVM_HC_DECLEAR_IN_0(__VA_ARGS__)
+#define PKVM_HC_DECLEAR_IN_2(...)	PKVM_HC_DECLEAR_IN_1(__VA_ARGS__)
+#define PKVM_HC_DECLEAR_IN_3(...)	PKVM_HC_DECLEAR_IN_2(__VA_ARGS__)
+#define PKVM_HC_DECLEAR_IN_4(...)	PKVM_HC_DECLEAR_IN_3(__VA_ARGS__)
+#define PKVM_HC_DECLEAR_IN_5(...)	PKVM_HC_DECLEAR_IN_4(__VA_ARGS__)
```
vineethrp: We could probably have empty defines for PKVM_HC_DECLARE_{IN,OUT}_{0-5} ?

arch/x86/include/asm/kvm_pkvm.h
```diff
@@ -172,22 +172,105 @@ static inline int pkvm_hc_input_num(enum pkvm_hc hc)
...
+#define PKVM_HC_DECLEAR_IN_6(...)	PKVM_HC_DECLEAR_IN_5(__VA_ARGS__);		\
```
vineethrp: As you suggested before, do we need to wrap code for args 6 to 10 in CONFIG_X86_64? I had it in my patch I shared before.

arch/x86/include/asm/kvm_pkvm.h
```diff
@@ -303,11 +439,35 @@ DEFINE_PKVM_HC_OUTPUT(1, RBX)
 DEFINE_PKVM_HC_OUTPUT(2, RCX)
 DEFINE_PKVM_HC_OUTPUT(3, RDX)
 DEFINE_PKVM_HC_OUTPUT(4, RSI)
+DEFINE_PKVM_HC_OUTPUT(5, RDI)
+DEFINE_PKVM_HC_OUTPUT(6, R8)
+DEFINE_PKVM_HC_OUTPUT(7, R9)
+DEFINE_PKVM_HC_OUTPUT(8, R10)
+DEFINE_PKVM_HC_OUTPUT(9, R11)
```
vineethrp, line 444: Read somewhere that R11(and probably R12) might be considered special in x86_64 syscall ABI. Not sure if pKVM considers it special?

arch/x86/include/asm/kvm_pkvm.h
cxdong: Yes, I will remove that.

arch/x86/include/asm/kvm_pkvm.h
cxdong: What I suggest was to make CONFIG_PKVM_X86 depending on x86_64 so that we don't need to put CONFIG_X86_64 in the code, unless pKVM needs to support 32bit mode?

arch/x86/include/asm/kvm_pkvm.h
cxdong: I think for pkvm, we don't have such special requirement, as long as the host itself allows to modify R11 or R12 when executing vmcall instruction.

cxdong: With this patch, we can imagine the code when we wants to more than 10 parameters. Mostly we can support 13 parameters by consuming all the registers. Not sure if pvIOMMU will requirement more? If pvIOMMU still requirement more, then may need to use shared memory page to transfer the inputs and outputs. If we have to do that in the end, then maybe it is not necessary to extend the pkvm_hypercall to support so many parameters.

maluka-dmytro: In v6.12 we've been using a shared page for pvIOMMU (we've been reusing pvVMCS's `pv_param` page for it), and we realized that it required some extra complexity and caution:
1. Security: when the host passes some params in this shared page, pKVM must take care to not use those params "in place" while handling the hypercall, since a malicious host may modify those params on another CPU in the meantime, leading to unpredictable behavior. So pKVM needs to copy them to local variables first, and only after that, validate them if needed, and use them.
2. Although the shared page is per CPU, a pvIOMMU hypercall may be issued by the IOMMU driver's IRQ handler, and this IRQ may be triggered while some other code is about to issue another hypercall to pKVM using the shared page on the same CPU, so there is a race between both accessing the same shared page. So we need to disable IRQs before using the shared page for passing params to pKVM.

In v6.12 we quickly addressed both these issues, but that made the code more complicated and ugly than it could be, with quite some boilerplate code. So we've been hoping that passing any params in registers would help us to avoid all that trouble.

cxdong: I see. Thanks for the details.

cxdong:
Integrated to https://github.com/intel-staging/pKVM-IA/pull/82, close this one.

PR #65, opened 2026-01-23
SQUASHME: pKVM: x86: Fix the unexpected pKVM hypercall inputs

cxdong:
When passing for example a bool as the pKVM hypercall's input in the RBX register, the compiler may only fill the low bits of the RBX while keeps its high bits unchanged. However the pKVM hypervisor gets the input parameters from the entire 64bit registers, which can get unexpected value. To fix this, zero extend all the input arguments to make sure they can full fill the entire 64bit registers.

This can fix the kselftest kvm debug_regs test.

Fixes: 289beb3a4ede ("pKVM: x86: Define pkvm_hypercall macro to as the PV interface")

### DIFF
--- a/arch/x86/include/asm/kvm_pkvm.h
+++ b/arch/x86/include/asm/kvm_pkvm.h
@@ -167,10 +167,10 @@ static inline int pkvm_hc_input_num(enum pkvm_hc hc)
 }
 
 #define PKVM_HC_IN_0()
-#define PKVM_HC_IN_1(a1)		, "b"(a1)
-#define PKVM_HC_IN_2(a1, a2)		PKVM_HC_IN_1(a1), "c"(a2)
-#define PKVM_HC_IN_3(a1, a2, a3)	PKVM_HC_IN_2(a1, a2), "d"(a3)
-#define PKVM_HC_IN_4(a1, a2, a3, a4)	PKVM_HC_IN_3(a1, a2, a3), "S"(a4)
+#define PKVM_HC_IN_1(a1)		, "b"((unsigned long)a1)
+#define PKVM_HC_IN_2(a1, a2)		PKVM_HC_IN_1(a1), "c"((unsigned long)a2)
+#define PKVM_HC_IN_3(a1, a2, a3)	PKVM_HC_IN_2(a1, a2), "d"((unsigned long)a3)
+#define PKVM_HC_IN_4(a1, a2, a3, a4)	PKVM_HC_IN_3(a1, a2, a3), "S"((unsigned long)a4)
 
 #define PKVM_HC_OUT_0(o)
 #define PKVM_HC_OUT_1(o)		, "=b"((o)->raw.data[0])


maluka-dmytro:
> When passing for example a bool as the pKVM hypercall's input in the RBX register, the compiler may only fill the low bits of the RBX while keeps its high bits unchanged. However the pKVM hypervisor gets the input parameters from the entire 64bit registers, which can get unexpected value.

But if the host passes a bool, the PV interface implementation in pKVM should also take a bool, i.e. ignore the higher bits?

With which PV interface was the issue observed?

> This can fix the kselftest kvm debug_regs test.

IIRC this test was actually succeeding? In what case was it failing?

cxdong:
The PV interface is `refresh_apicv_exec_ctrl`. This is the asm code compiled by clang from my side:
```
    2130:       48 8b 87 18 03 00 00    mov    0x318(%rdi),%rax
    2137:       80 b8 98 00 00 00 00    cmpb   $0x0,0x98(%rax)
    213e:       0f 95 c3                setne  %bl
    2141:       b8 31 00 00 00          mov    $0x31,%eax
    2146:       0f 01 c1                vmcall
```

The `setne %bl` just set the low byte of rbx, thus the entire rbx may be a non-zero value, although we want to pass a false to this hypercall. Then `pkvm_hc_input1()` will get a non-zero unsigned long value, and converting this value to bool will get a true, which is different with what we want.

With gcc, the asm code  seems fine, `movzbl` will zero extended the value to the entire rbx.
```
    1e67:       0f b6 98 98 00 00 00    movzbl 0x98(%rax),%ebx
    1e6e:       b8 31 00 00 00          mov    $0x31,%eax
    1e73:       0f 01 c1                vmcall
```

maluka-dmytro:
> Then `pkvm_hc_input1()` will get a non-zero unsigned long value, and converting this value to bool will get a true

Aaaa, got it. Yeah, when converting u64 to bool, all bits are taken into account (unlike e.g. when converting u64 to u8, in which case the higher bits would be just ignored so there would be no issue). Great catch.

maluka-dmytro (approved):

maluka-dmytro:
LGTM. I guess we will merge this patch as is, without squashing it into 289beb3 (since 289beb3  is already merged in Android kernel)? i.e. at most we can reorder it right after the pvVMCS-part1 patches?

cxdong:
> LGTM. I guess we will merge this patch as is, without squashing it into [289beb3](https://github.com/intel-staging/pKVM-IA/commit/289beb3a4ede4df762faac6171e1a5947ae5482e) (since [289beb3](https://github.com/intel-staging/pKVM-IA/commit/289beb3a4ede4df762faac6171e1a5947ae5482e) is already merged in Android kernel)? i.e. at most we can reorder it right after the pvVMCS-part1 patches?

Yes, we will merge it as it is. The SQUASHME just reminds us to squash it when we rebase pKVM to some new version of kernel, so we don't even to reorder it but just apply it on top of the current tree. If SQUASHME is confusing, maybe we can replace it with a "FIXME", so that this can also reminds us this is a fix patch just from the patch title rather than looking into the patch commit?

maluka-dmytro:
SQUASHME sounds ok to me. (FIXME would rather suggest that this patch itself introduces an issue which needs to be fixed, not that it fixes it?)

PR #66, opened 2026-02-02
Pkvm v6.18: pvMMU + pvmfw support + pvSIPI

maluka-dmytro:
Port the following pKVM-IA functionality to v6.18:
* guest MMU support (also known as pvMMU)
* pvmfw support
* secure startup of secondary vCPUs (also known as pvSIPI)


arch/x86/include/asm/kvm_host.h
```diff
@@ -849,12 +851,39 @@ static inline void *pop_pkvm_memcache_page(struct pkvm_memcache *mc,
...
 
+/* For memcaches containing single-page ranges only. */
+static inline int topup_pkvm_memcache(struct pkvm_memcache *mc,
+				      unsigned long min_pages,
+				      void *(*alloc_pg)(void *arg),
+				      phys_addr_t (*to_pa)(void *virt),
+				      void *arg)
+{
+	while (mc->count < min_pages) {
+		phys_addr_t *p = alloc_pg(arg);
```
cxdong: nit: seems we want to ephasize that the allocated memory will be used to store the type of `phys_addr_t`? Or we just use void *?

arch/x86/include/asm/kvm_host.h
```diff
@@ -788,6 +788,8 @@ struct pkvm_memcache {
 		u64 nr_pages;
 	} head;
 	unsigned long count;
+#define PKVM_MC_ACCOUNT_PGTABLE_PAGES	BIT(1)
+	unsigned long flags;
```
cxdong, line 793: Not sure if we ever discussed about this. Currently this flags is for the entire `pkvm_memcache`, so all the page ranges in this `pkvm_memcahce` should have the same flags. And this looks like is the reason that, we need to define a `struct pkvm_memcache guest_mmu_teardown_mc;` in the `struct kvm_pkvm_vm`. If so, seems putting flags to `struct pkvm_page_range` can make it as a per range flag, so that both the guest mmu pages and vm/vcpu structure pages can be in the same `pkvm_memcache`. I am not suggesting we should do that, just want to see if I have missed something.

arch/x86/kvm/pkvm/memory.h
```diff
@@ -140,6 +140,16 @@ static inline void pkvm_clear_memory(void *va, size_t size)
...
+static inline void *pkvm_phys_to_virt(phys_addr_t phys)
+{
+	return __pkvm_va(phys);
+}
+
+static inline phys_addr_t pkvm_virt_to_phys(void *addr)
+{
+	return __pkvm_pa(addr);
+}
+
```
cxdong, line 174: nit: the commit message "s/They will we used/They will be used"?

arch/x86/include/asm/kvm_host.h
```diff
@@ -895,6 +895,7 @@ struct kvm_pkvm_vm {
 
 struct kvm_pkvm_vcpu {
 	int handle;
+	struct pkvm_memcache guest_mmu_memcache;
```
cxdong, line 938: Just want to understand the reason of choosing this direction (pKVM-ARM is doing the same). It seems the reason is that the `pkvm_pool` cannot be extended dynamically, and also cannot support non-contigous pages, so allocating guest mmu page table pages from the `pkvm_memcache` is easier, and may be more efficient as it is a per-vCPU thing which doesn't need synchronizations for SMP?

arch/x86/kvm/pkvm/mmu.c
```diff
@@ -579,6 +579,27 @@ void pkvm_guest_mmu_destroy(struct pkvm_vm *pkvm_vm)
...
+	while (vcpu_mc->count) {
+		/* Drain hyp owned memcache and push pages to the teardown memcache */
+		addr = pop_pkvm_memcache_page(vcpu_mc, pkvm_phys_to_virt);
+
+		/*
+		 * Since pages comes from non-used memcache, there is no need to
+		 * zero them before pushing to teardown_mc (which host can
+		 * access after donating them back to the host in next step).
+		 */
+		push_pkvm_memcache_page(teardown_mc, addr, pkvm_virt_to_host_gpa);
```
cxdong: The teardow_mc is per-vm, which is accessed without holding a lock. So the vcpu should be freed by the host one by one to make sure this won't happen concurrently on different CPUs. And this is also how the host KVM does today. If so, maybe worth to mention this requirement in the above comment?

arch/x86/kvm/pkvm/mmu.c
```diff
@@ -403,6 +403,36 @@ static void drain_pool(struct pkvm_pool *pool, struct pkvm_memcache *host_mc)
...
+	page = pop_pkvm_memcache_page(host_mc, pkvm_host_gpa_to_virt);
+	if (!page)
+		return NULL;
+
+	/*
+	 * The page will be cleared by zalloc_page when allocating it from
+	 * the memcache, thus no need to clear it now.
+	 */
+	if (pkvm_host_donate_hyp(__pkvm_pa(page), PAGE_SIZE, false)) {
+		WARN_ON(1);
```
cxdong: nit: maybe just `if (WARN_ON(pkvm_host_donate_hyp(__pkvm_pa(page), PAGE_SIZE, false)))`?

arch/x86/kvm/pkvm/mmu.c
```diff
@@ -937,3 +980,66 @@ void pkvm_host_unshare_hyp(unsigned long phys, unsigned long size)
...
+{
+	u64 prot = guest_mmu_pte_prot(vcpu, gpa, true, PKVM_PAGE_OWNED);
+	struct pkvm_vm *pkvm_vm = to_pkvm_vcpu(vcpu)->pkvm_vm;
+	int ret;
+
+	if (!PAGE_ALIGNED(gpa) || !PAGE_ALIGNED(hpa) ||
+	    !PAGE_ALIGNED(size) || size == 0)
+		return -EINVAL;
+
+	if (WARN_ON_ONCE(!pkvm_is_protected_vm(&pkvm_vm->kvm)))
```
cxdong: nit: it seems using `pkvm_is_protected_vcpu(vcpu)` can make the code a bit shorter?

arch/x86/kvm/pkvm/mmu.c
```diff
@@ -16,6 +17,9 @@ pkvm_spinlock_t host_mmu_lock;
 static const struct pkvm_pgtable_ops *guest_mmu_pgt_ops;
 static struct pkvm_pgtable_cap guest_mmu_pgt_cap;
 
+DEFINE_PER_CPU(struct pkvm_vm *, __current_vm);
```
cxdong: nit: maybe can add static if `__current_vm` is not used outside this file?

arch/x86/kvm/pkvm/mmu.c
```diff
@@ -1043,3 +1046,87 @@ int pkvm_host_donate_guest(struct kvm_vcpu *vcpu, unsigned long gpa,
...
+			  bool writable)
+{
+	u64 prot = guest_mmu_pte_prot(vcpu, gpa, writable,
+				      PKVM_PAGE_SHARED_BORROWED);
+	struct pkvm_vm *pkvm_vm = to_pkvm_vcpu(vcpu)->pkvm_vm;
+	int ret;
+
+	if (!PAGE_ALIGNED(gpa) || !PAGE_ALIGNED(hpa) ||
+	    !PAGE_ALIGNED(size) || size == 0 ||
+	    !is_memory_range(hpa, size))
```
cxdong, line 1354: `pkvm_host_donate_guest` doesn't have the check `!is_memory_range(hpa, size)`, shall we also add this check to  `pkvm_host_donate_guest` or any consideration for not doing that?

arch/x86/kvm/mmu/mmu.c
```diff
@@ -4914,12 +4917,79 @@ static int kvm_tdp_mmu_page_fault(struct kvm_vcpu *vcpu,
...
+	r = kvm_mmu_faultin_pfn(vcpu, fault, ACC_ALL);
+	if (r != RET_PF_CONTINUE) {
+		/* MMIO emulation works for non-protected VMs only. */
+		if (unlikely(pkvm_is_protected_vcpu(vcpu) && r == RET_PF_EMULATE))
+			return -EFAULT;
+
+		if (unlikely(r != RET_PF_EMULATE && r != RET_PF_RETRY &&
+			     r != -EINTR && r != -EAGAIN && r != -EFAULT)) {
+			pr_warn_ratelimited("pkvm: unexpected page fault result %d\n", r);
+			WARN_ON_ONCE(1);
```
cxdong, line 5028: nit: maybe `if(unlikely(WARN_ON_ONCE(...)))`?

arch/x86/kvm/mmu/mmu.c
```diff
@@ -4954,6 +4968,11 @@ static int pkvm_page_fault(struct kvm_vcpu *vcpu, struct kvm_page_fault *fault)
 	if (r)
 		return r;
 
+	/* Allocate non-atomically before taking mmu_lock. */
+	mapping = kzalloc(sizeof(struct pkvm_mapping), GFP_KERNEL_ACCOUNT);
```
cxdong, line 5041: Seems you fixed an issue in v6.12, which does non-atomic allocation after taking the lock.

arch/x86/include/asm/kvm_host.h
```diff
@@ -886,11 +887,44 @@ static inline void init_pkvm_mmu_memcache(struct pkvm_memcache *mc)
...
+ * Caution: __start and __end are evaluated multiple times.
+ */
+#define for_each_pkvm_mapping(__kvm, __start, __end, __map)					\
+	for (struct pkvm_mapping *__tmp = pkvm_mapping_iter_first(&(__kvm)->arch.pkvm.mappings,	\
+								  (__start), (__end) - 1);	\
+	     __tmp && ({									\
+				__map = __tmp;							\
+				__tmp = pkvm_mapping_iter_next(__map, (__start), (__end) - 1);	\
+				true;								\
+		       });									\
```
cxdong, line 921: nit: the code in `({})` seems should be the 3rd part of the for loop? Maybe it can be changed to:
```
  #define for_each_pkvm_mapping(__kvm, __start, __end, __map)                                     \
          for (struct pkvm_mapping *__tmp = pkvm_mapping_iter_first(&(__kvm)->arch.pkvm.mappings, \
                                                                    (__start), (__end) - 1);      \
               __tmp; ({                                                                          \
                                  __map = __tmp;                                                  \
                                  __tmp = pkvm_mapping_iter_next(__map, (__start), (__end) - 1);  \
                       })                                                                         \
              )
```
and also doesn't need to explicitly add a `true;`?

arch/x86/kvm/mmu/mmu.c
```diff
@@ -6838,6 +6865,11 @@ int kvm_mmu_init_vm(struct kvm *kvm)
 {
 	int r, i;
 
+#ifdef CONFIG_PKVM_X86
+	if (enable_pkvm)
+		kvm->arch.pkvm.mappings = RB_ROOT_CACHED;
```
cxdong, line 6982: nit: does it look better if we move this to the `pkvm_vm_init` in the pkvm_host.c? Then it seems more consistent with the destroy part, which is also done by the `pkvm_vm_destroy` in the pkvm_host.c, and also can avoid change the upstream code.

arch/x86/include/asm/kvm_host.h
jaszczyk-grzegorz: I think we could, but maybe it is worth keeping as it to be aligned with pKVM-ARM memchaces where they also use it that way. Also in 6.12 `push_pkvm_memcache` we used `phys_addr_t` but seems like @maluka-dmytro has changed it in `push_pkvm_memcache_page` to use void*. BTW seems like he also changed the name of  __topup_pkvm_memcache to topup_pkvm_memcache, not sure why. I am ok with both but as we discussed earlier my intention was to be close to what ARM is doing since someday maybe we will want to unify memcaches for both pKVM-ARM and pKVM-IA but yeah I remember we already moved a bit from the original pKVM-ARM API and naming

arch/x86/include/asm/kvm_host.h
jaszczyk-grzegorz: Yeah I think what you wrote is correct. This was ported from 6.12 where we didn't had `pkvm_page_range` similar as it was (in 6.12) and is handled by pKVM-ARM in 6.18

arch/x86/kvm/pkvm/mmu.c
jaszczyk-grzegorz: in 6.12 they were used in two files but for 6.18 I think we can change it to static

arch/x86/include/asm/kvm_host.h
maluka-dmytro: > I think we could, but maybe it is worth keeping as it to be aligned with pKVM-ARM memchaces where they also use it that way.

Actually I think `phys_addr_t *` makes no sense here, so I'll probably change it to `void *`. It made some sense in pKVM-ARM where a `phys_addr_t` address is stored in this memory, but in our implementation here we don't store `phys_addr_t` there, we store `struct pkvm_page_range` there instead.

arch/x86/include/asm/kvm_host.h
jaszczyk-grzegorz: There are also some details that we discussed with @maluka-dmytro when it was reviewed in 6.12 e.g.: https://android-review.git.corp.google.com/c/kernel/common/+/3702531/comment/0f1ce9ea_412842c1/ . The pool is still used and extended in rare case let me quote:
"They are put to the per-VM pool quite rarely: note that guest_mmu_put_page actually call __hyp_put_page which puts the page to the pool only when the page is not used anymore (hyp_page_ref_dec_and_test(p)), which happen mostly during VM teardown (could happen due to MMU notifiers, e.g. when ballooning jumps in but even then it is rare that all PTE for a given page-table page (all 512 PTE) are not used and therefore page table page is freed and put to the pool)..."

I think there are two reason why to do it that way in general (despite the fact that it was based on pKVM-ARM):
1) per-vCPU memcache doesn't need synchronization for SMP
2) the guest_mmu_zalloc_page could be written differently to always allocate pages from the per-VM poll but that would mean that  when feeling host backed pages we will have to be put them into the per-vm poll in the first place so there will be unneeded overhead of attaching the page to the pool in hyp_put_page() -> __hyp_attach_page() every time we need to refill it.

arch/x86/include/asm/kvm_host.h
maluka-dmytro: We did discuss this (I mean, all three of us did :)). And I was suggesting precisely the same: the flag could be a per-range flag, and then we wouldn't need the `guest_mmu_teardown_mc` field, as we could reuse the teardown memcache passed via the hypercall args. We concluded that functionality-wise it would have no advantages (at least for now), while it would introduce even more discrepancy from pKVM-ARM, so we ended up not doing that.

arch/x86/kvm/pkvm/memory.h
maluka-dmytro: Yep, thanks.

arch/x86/include/asm/kvm_host.h
maluka-dmytro: If you are asking why we are using a memcache for this at all, the answer is simple: yes, we want to allocate those pages dynamically in the host and let the host dynamically provide them to pKVM (so that we don't need to reserve a large pKVM memory pool for allocating them in pKVM; so the size of pKVM's reserved memory is much smaller and doesn't need to depend on the max number of VMs). And the memcache is a convenient mechanism for providing pages from the host to pKVM (since the pKVM memcache API is simple and can be used both by the host and by the hypervisor,  unlike the pkvm_pool API which is just for the hypervisor).

> It seems the reason is that the `pkvm_pool` cannot be extended dynamically

To be precise, actually it can (it's a non-obvious trick): we can add an "external" page to a pool via `pkvm_put_page()`, thus dynamically extending it. And we are actually using that (though not most of the time): when we are freeing a page in `guest_mmu_put_page()`, we are putting this page into `current_vm->mmu_pool`, while this page, most probably, was not allocated from this pool, it was allocated from the memcache instead (in `guest_mmu_zalloc_page()`).

So maybe actually we could do that all the time, i.e.: instead of having two instances of `kvm_pkvm_vcpu->guest_mmu_memcache` (the host's one and the hypervisor's one) we could have just the host's one, while on the hypervisor side, after popping a page from the host memcache, we could just put it into the `pkvm_vm->mmu_pool` right away, instead of pushing it to the hypervisor memcache (IOW we could have "pool refill" instead of "memcache refill"), so that pKVM always allocates from the pool afterwards, not from the memcache. That would be probably just slightly less efficient (since: 1. allocating a page from the pool via pKVM's buddy allocator is probably slower than allocating it from the memcache. 2. attaching a page to the pool during "pool refill" is probably also slower than just pushing it to the memcache. 3. we would need to do "pool refill" after taking the `pkvm_vm->mmu_lock`, not before.). And it seems that would have no advantages either. So we just did what pKVM-ARM did.

> and also cannot support non-contigous pages

I'm not sure why you are mentioning this. pkvm_memcache doesn't support non-contiguous pages either.

arch/x86/kvm/pkvm/mmu.c
maluka-dmytro: Yep, we can. @jaszczyk-grzegorz I assume there was no particular reason for WARN_ON(1) here instead?

arch/x86/kvm/pkvm/mmu.c
maluka-dmytro: Yep, good point.

arch/x86/kvm/pkvm/mmu.c
maluka-dmytro: Yeah, right. I didn't realize we could do `static DEFINE_PER_CPU`.

arch/x86/kvm/pkvm/mmu.c
maluka-dmytro: `pkvm_host_donate_guest()` has the check `check_host_mem_pgstate()` which itself includes a `is_memory_range()` check. While `pkvm_host_share_guest()` doesn't check `check_host_mem_pgstate()` so it needs to explicitly check `is_memory_range()`.

BTW it's the same situation as with `pkvm_host_donate_hyp()` vs `pkvm_host_share_hyp()`.

arch/x86/kvm/mmu/mmu.c
maluka-dmytro: Let me just copy my answer from https://android-review.googlesource.com/c/kernel/common/+/3666641/17/arch/x86/kvm/mmu/mmu.c#4740 :)

The reason I did it this way is just to have this error message above the call trace, not below it.

arch/x86/kvm/mmu/mmu.c
maluka-dmytro: Yeah, exactly. BTW @vineethrp at least once reproduced an issue which was probably caused by this, during pKVM-IA stress testing on v6.12: a soft lockup with a call trace indicating that `pkvm_pin_page()` triggered `__alloc_pages_direct_reclaim()` thus took a lot of time (e.g. due to swapping or whatever).

We could backport this fix to v6.12 if needed, but I hope there is not much need, since that 6.12 branch is gonna be deprecated soon.

BTW I've just tried enabling CONFIG_DEBUG_ATOMIC_SLEEP=y and moving this kzalloc after taking the lock, and easily reproduced multiple call traces caused by it. I guess we never tried enabling it on v6.12 which is why we didn't easily notice this issue.

...BTW instead of this kzalloc we could introduce a KVM memcache specifically for these `struct pkvm_mapping` allocations, i.e. similar to regular KVM's `mmu_page_header_cache`, `pte_list_desc_cache` etc. That would be probably more efficient, but that would mean a bit more changes in the upstream code, so I decided to simply use kzalloc here for now.

arch/x86/include/asm/kvm_host.h
maluka-dmytro: Just noticed that Greg has replied before I did. Yeah, we also discussed this a bit in https://android-review.googlesource.com/c/kernel/common/+/3702531/comment/0f1ce9ea_412842c1/ , thanks for reminding me about that discussion.

arch/x86/include/asm/kvm_host.h
maluka-dmytro: At first I thought "yeah, seems so", but I've just tried that, and it crashes, at the very first iteration of `for_each_pkvm_mapping()`.  I believe the reason is that it makes `__map` uninitialized until the 2nd iteration (while the loop body, obviously, already tries to do something with `__map` during the 1st iteration).

BTW this implementation of `for_each_pkvm_mapping()` is basically just copied from pKVM-ARM's `for_each_mapping_in_range_safe()`.

arch/x86/include/asm/kvm_host.h
maluka-dmytro: > BTW seems like he also changed the name of __topup_pkvm_memcache to topup_pkvm_memcache, not sure why.

That is to keep it consistent with the naming style we've been already using for other memcache functions in v6.18. For instance, we have `free_pkvm_memcache()` (used by both host and pKVM) here in kvm_host.h, and its wrapper `kvm_free_pkvm_memcache()` (used by host only) in pkvm.c. Thus, similarly we have `topup_pkvm_memcache()` and `kvm_topup_pkvm_memcache()`.

arch/x86/kvm/mmu/mmu.c
maluka-dmytro: Actually it seems cleaner to me to do this here in mmu.c, since this is directly related to MMU, plus this is not VMX-specific. Also please note that this is not the only `if (enable_pkvm)` check here, we have another one below (to avoid needlessly calling `kvm_mmu_alloc_page_hash()`) as well as in `kvm_mmu_create()` and `kvm_mmu_destroy()`.

As for destroying it in `pkvm_vm_destroy()`, I actually added a comment about it there:
```
	/*
	 * It would be nice to do this in generic code in kvm_mmu_uninit_vm(),
	 * not here. But kvm_mmu_uninit_vm() is called too early, when pKVM
	 * has not released guest pages to the host yet.
	 */
```

so I see it rather as a temporary solution, BTW related to the fact that currently we completely postpone guest MMU teardown by pKVM until the moment of vm_destroy, which is actually problematic in itself, since we are spending really a lot of time (e.g. hundreds of millions of cycles for a pVM) handling the vm_destroy hypercall in pKVM. So we might eventually rework it similarly to pKVM-ARM: let the host initiate _starting_ VM teardown (which means just letting pKVM forbid running VM's vCPUs starting from that point) and then let the host reclaim its pages one by one via separate hypercalls, before the actual VM teardown via vm_destroy in the end.

arch/x86/kvm/pkvm/mmu.c
maluka-dmytro: Yeah, makes sense. Or maybe better to add such a comment before calling `pkvm_guest_mmu_free_memcache()` in `__pkvm_vcpu_free()` (so that it is clear from the context why we are talking about vCPU freeing)?

maluka-dmytro: > Or maybe better to add such a comment before calling `pkvm_guest_mmu_free_memcache()` in `__pkvm_vcpu_free()` (so that it is clear from the context why we are talking about vCPU freeing)?

I've ended up adding it here as you suggested, not in `__pkvm_vcpu_free()`, to keep the `guest_mmu_teardown_mc` an "implementation detail" of pkvm/mmu.c.

arch/x86/kvm/mmu/mmu.c
maluka-dmytro: I'll slightly modify that comment in `pkvm_vm_destroy()` to make it a TODO.

arch/x86/kvm/pkvm/mmu.c
jaszczyk-grzegorz: Yeah, I do not recall any particular reason for above and I think it is better to change as suggested (and as you already did:) )

arch/x86/include/asm/kvm_host.h
cxdong: Thanks for the explainations. I didn't realize the `pkvm_pool` can be extended dynamically with non-contigous pages. The reasons you guys explained make sense. Using a per-vcpu `pkvm_memcache` is more efficient than extending and allocating from the `pkvm_pool`.

arch/x86/include/asm/kvm_host.h
cxdong: I see. Thanks!

arch/x86/kvm/mmu/mmu.c
cxdong: Ah, sorry, I forget about that. :)

arch/x86/kvm/mmu/mmu.c
cxdong: > ...BTW instead of this kzalloc we could introduce a KVM memcache specifically for these struct pkvm_mapping allocations, i.e. similar to regular KVM's mmu_page_header_cache, pte_list_desc_cache etc. That would be probably more efficient, but that would mean a bit more changes in the upstream code, so I decided to simply use kzalloc here for now.

Yeah, makes sense.

arch/x86/include/asm/kvm_host.h
cxdong: I see. I missed the truth that __map is used by the 1st iteration. Below works from my side:
```
  #define for_each_pkvm_mapping(__kvm, __start, __end, __map)                                     \
          for (struct pkvm_mapping *__tmp = pkvm_mapping_iter_first(&(__kvm)->arch.pkvm.mappings, \
                                                                    (__start), (__end) - 1);      \
               (__map = __tmp);                                                                   \
               __tmp = pkvm_mapping_iter_next(__map, (__start), (__end) - 1)                      \
              )
```
This seems more simpler but not sure if this is more readble than the current version as it use one single expression `(__map = __tmp)` to assign `__map` and check `__tmp` is NULL or not?

arch/x86/kvm/mmu/mmu.c
cxdong: I see. Makes sense. Thanks!

arch/x86/kvm/mmu/mmu.c
```diff
@@ -5006,6 +5006,25 @@ static int pkvm_page_fault(struct kvm_vcpu *vcpu, struct kvm_page_fault *fault)
 	mapping->gfn = base_gfn;
 	mapping->pfn = fault->pfn;
 	mapping->nr_pages = nr_pages;
+
+	if (pkvm_is_protected_vcpu(vcpu)) {
+		if (WARN_ON_ONCE(!fault->refcounted_page)) {
+			r = -EFAULT;
```
cxdong, line 5081: Just a note: If we use this function to handle the EPT violation of accessing a passthrough device MMIO for the pVM in the future, then seems the `fault->refcounted_page` will be NULL which is reasonable? If so, then seems both `!fault->refcounted_page` and `fault->refcounted_page` are reasonable. Then maybe we just don't pin the page if `!fault->refcounted_page`?

arch/x86/kvm/pkvm/mmu.c
```diff
@@ -493,6 +493,104 @@ static int refill_memcache(struct pkvm_memcache *mc, unsigned long min_pages,
...
+static int host_reclaim_guest_walker(struct pkvm_pgtable_visit_ctx *ctx,
+				     unsigned long walk_flags,
+				     void *const arg)
+{
+	const struct pkvm_pgtable_ops *pgt_ops = ctx->pgt->pgt_ops;
+	struct kvm *kvm = pgt_to_kvm(ctx->pgt);
+	unsigned long phys, size;
+	void *ptep = ctx->ptep;
+
+	if (!pgt_ops->pte_present(ptep) && !pgt_ops->pte_annotated(ptep))
```
cxdong: If we annotate guest mmu pte, means this pte doesn't map to a physical page but contains some annotation. In this case,  `pgt_ops->pte_annotated(ptep)` is true and we will continue rather than `return 0;`. But seems below code only handle for the present pte rather than for the annotated pte? Or we expect the `default` case should happen for the annotated pte? BTW, do we have the scenario to annotate the guest mmu pte? If no, maybe can just `BUG_ON(pgt_ops->pte_annotated(ptep))`?

arch/x86/kvm/pkvm/mmu.c
```diff
@@ -591,6 +591,82 @@ static void host_reclaim_guest_pages(struct pkvm_vm *pkvm_vm)
...
+				  enum pkvm_page_state host_state)
+{
+	int ret;
+
+	ret = check_page_state(&pkvm_vm->mmu, gpa, size, guest_state);
+	if (ret)
+		return ret;
+
+	return for_each_contig_range(&pkvm_vm->mmu, gpa, size,
+				     __check_guest_host_state, &host_state);
```
cxdong, line 650: It seems we can also check the guest and host page state in one guest mmu page table walk, e.g., implement a walker to check both for a present guest mmu pte. Then don't need to use `pkvm_pgtable_lookup_range` to walk the guest mmu again to get the physical address. Or maybe I missed something that we have to use this way?

arch/x86/kvm/pkvm/pgtable.c
```diff
@@ -913,3 +935,34 @@ void pkvm_pgtable_destroy(struct pkvm_pgtable *pgt)
...
 }
+
+/**
+ * pkvm_pgtable_test_clear_young() - Test and optionally clear the access flag
+ *				     in the page table entries for the given
+ *				     virtual address range.
+ * @pgt:	The page table.
+ * @vaddr:	The start address of virtual address range.
+ * @size:	The size of the virtual address range.
+ * @mkold:	If true, clear the access flag if it is set.
```
cxdong, line 946: nit: s/if it is set//, as we already have `If true,` at the beginning?

arch/x86/kvm/pkvm/mmu.c
```diff
@@ -1351,3 +1351,43 @@ int pkvm_host_unshare_guest(struct kvm *kvm, unsigned long gpa,
...
+/**
+ * pkvm_host_test_clear_young_guest() - Test and optionally clear the access
+ *					flag in guest PTEs for host pages
+ *					shared with a guest.
+ * @kvm:	Guest VM.
+ * @gpa:	Address of the guest physical address region to test/clear the
+ *		access flag.
+ * @size:	Size of the guest physical address region to test/clear the
+ *		access flag.
+ * @mkold:	If true, clear the access flag if it is set.
```
cxdong, line 1459: nit: s/if it is set//? Similar to the last patch.

arch/x86/kvm/vmx/main.c
```diff
@@ -978,8 +978,10 @@ struct kvm_x86_ops vt_x86_ops __initdata = {
 
 	.set_tss_addr = vt_op(set_tss_addr),
 	.set_identity_map_addr = vt_op(set_identity_map_addr),
+#endif
 	.get_mt_mask = vmx_get_mt_mask,
```
cxdong, line 982: It seems we need to enforce enable_ept in the host side, otherwise the host can still set enable_ept as 0, which will set this ops as NULL by below code in vmx_hardware_setup:
```
  #ifndef __PKVM_HYP__
          if (enable_ept)
                  kvm_mmu_set_ept_masks(enable_ept_ad_bits,
                                        cpu_has_vmx_ept_execute_only());
          else
                  vt_x86_ops.get_mt_mask = NULL;

          /*
           * Setup shadow_me_value/shadow_me_mask to include MKTME KeyID
           * bits to shadow_zero_check.
           */
          vmx_setup_me_spte_mask();

          kvm_configure_mmu(enable_ept, 0, vmx_get_max_ept_level(),
                            ept_caps_to_lpage_level(vmx_capability.ept));
  #else
          pkvm_guest_ept_setup();
  #endif
```

arch/x86/kvm/pkvm/mmu.c
cxdong: I see `for_each_contig_range` is also used by other code. Then seems implementing a new walker is making code more complex if `for_each_contig_range` is inevitable, and reusing `for_each_contig_range` is easier, although it introduced additional page table walk, the additional walk is not in a critial path.

arch/x86/kvm/pkvm/mmu.c
```diff
@@ -667,6 +667,44 @@ static int __host_unshare_guest(unsigned long gpa, unsigned long hpa,
...
+
+	BUG_ON(pkvm_pgtable_set_owner(&host_mmu, hpa, size, PKVM_ID_GUEST));
+	set_host_mem_pgstate(hpa, size, PKVM_PAGE_NONE);
+
+	prot = pkvm_pte_set_pgstate(prot, &pkvm_vm->mmu, PKVM_PAGE_OWNED);
+	ret = pkvm_pgtable_map(&pkvm_vm->mmu, gpa, hpa, size, prot, NULL);
+	if (ret)
+		return ret;
+
+	return 0;
```
cxdong: maybe just `return pkvm_pgtable_map(&pkvm_vm->mmu, gpa, hpa, size, prot, NULL);` and no need to have `ret`?

arch/x86/kvm/pkvm/mmu.c
```diff
@@ -667,6 +667,44 @@ static int __host_unshare_guest(unsigned long gpa, unsigned long hpa,
...
+{
+	struct kvm_vcpu *vcpu = arg;
+	struct pkvm_vm *pkvm_vm = to_pkvm_vcpu(vcpu)->pkvm_vm;
+	int ret;
+
+	BUG_ON(pkvm_pgtable_set_owner(&host_mmu, hpa, size, PKVM_ID_GUEST));
+	set_host_mem_pgstate(hpa, size, PKVM_PAGE_NONE);
+
+	prot = pkvm_pte_set_pgstate(prot, &pkvm_vm->mmu, PKVM_PAGE_OWNED);
+	ret = pkvm_pgtable_map(&pkvm_vm->mmu, gpa, hpa, size, prot, NULL);
```
cxdong: The `pkvm_pgtable_map` in `__guest_share_host` is taking `&vcpu->arch.pkvm.guest_mmu_memcache` as the `pkvm_memcache`, but here takes `NULL`. I guess the function should be fine as we just change the prot bits  for a present pte, thus it should not allocate any page table page, right? Maybe we can change one of them to make the usage consistent?

arch/x86/kvm/pkvm/mmu.c
```diff
@@ -1391,3 +1429,91 @@ int pkvm_host_test_clear_young_guest(struct kvm *kvm, unsigned long gpa,
...
+				     PKVM_PAGE_NONE);
+	if (ret)
+		goto unlock;
+
+	ret = for_each_contig_range(&pkvm_vm->mmu, gpa, size,
+				    __guest_share_host, vcpu);
+	if (ret) {
+		/* Unshare already shared pages, if any. */
+		for_each_contig_range(&pkvm_vm->mmu, gpa, size,
+				      __guest_unshare_host, vcpu);
```
cxdong: If `__guest_unshare_host` returns negative value, we just siliencely failed. Shall we use a BUG_ON for that?

arch/x86/kvm/pkvm/mmu.c
```diff
@@ -1391,3 +1429,91 @@ int pkvm_host_test_clear_young_guest(struct kvm *kvm, unsigned long gpa,
...
+	pkvm_guest_mmu_lock(pkvm_vm);
+
+	ret = check_guest_host_state(pkvm_vm, gpa, size, PKVM_PAGE_SHARED_OWNED,
+				     PKVM_PAGE_SHARED_BORROWED);
+	if (ret)
+		goto unlock;
+
+	ret = for_each_contig_range(&pkvm_vm->mmu, gpa, size,
+				    __guest_unshare_host, vcpu);
+	BUG_ON(ret);
```
cxdong, line 1570: nit: maybe just `BUG_ON(for_each_contig_range(&pkvm_vm->mmu, gpa, size, __guest_unshare_host, vcpu)` and `return 0` in the end?

BTW, it seems the pVM guest kernel doesn't expect to receive an error code from the hypercall PKVM_GHC_UNSHARE_MEM, otherwise seems we can return the error code to the guest via the RAX rather than BUG_ON?

arch/x86/kvm/pkvm/mmu.c
cxdong: I see in the next path, it mentioned we may need to split huge mappings by the `__guest_share_host` thus it has to use `&vcpu->arch.pkvm.guest_mmu_memcache` to allocate page table pages. And `__guest_unshare_host` doesn't need this so it can use NULL. We don't need to make the usage consistent. Please ignore the above comment.

arch/x86/kvm/x86.c
```diff
@@ -10653,6 +10669,15 @@ static int kvm_pkvm_hypercall(struct kvm_vcpu *vcpu)
...
 		ret = kvm_sev_es_mmio_write(vcpu, kvm_rbx_read(vcpu), size, &val);
 		break;
+	case PKVM_GHC_SHARE_MEM:
+		/*
+		 * The only case when pKVM forwards this hypercall to the host
+		 * is when it asks the host to refill the memcache with the
+		 * needed amount of pages.
+		 */
+		refill_size = vcpu->arch.pkvm.req_param;
+		ret = handle_memcache_refill(vcpu, refill_size);
```
cxdong: nit: maybe can simplify as below so that don't need to define `handle_memcache_refill`, and also saving the `unsigned long refill_size`?
```
                  ret = kvm_topup_pkvm_memcache(&vcpu->arch.pkvm.guest_mmu_memcache,
                                                (unsigned long)vcpu->arch.pkvm.req_param);
                  if (!ret)
                          ret = 1;
```

arch/x86/kvm/vmx/vmx.c
```diff
@@ -9765,14 +9765,16 @@ static void update_protected_vcpu_state(struct kvm_vcpu *vcpu,
 		to_pkvm_vcpu(vcpu)->host_emulated_msr_err = 0;
 		break;
 	case EXIT_REASON_VMCALL:
+		if (pkvm_hypercall_need_skip_instruction(kvm_rax_read(vcpu)))
+			WARN_ON_ONCE(kvm_skip_emulated_instruction(vcpu) != 1);
```
cxdong: For the `PKVM_GHC_SHARE_MEM`, we will not skip the vmcall instruction, so the guest will execute vmcall again with the hypercall number in the RAX. So if `kvm_topup_pkvm_memcache` is successful, seems we should set `PKVM_GHC_SHARE_MEM` in pVM's RAX to make sure the guest will retry this hypercall rather than something else? Currenly the pVM's RAX will be overriden by below code.

BTW, what if the `kvm_topup_pkvm_memcache` is failed?

arch/x86/kvm/x86.c
```diff
@@ -14959,6 +14989,24 @@ int pkvm_emulate_hypercall(struct kvm_vcpu *vcpu)
...
+		memcache_refill_size = __pkvm_pgtable_max_pages(a1 >> PAGE_SHIFT);
+		if (vcpu->arch.pkvm.guest_mmu_memcache.count < memcache_refill_size) {
+			/*
+			 * If not enough pages in the memcache to handle mapping, request
+			 * should be forwarded to the host for refilling the memcache with
+			 * the needed amount of pages, after which the instruction will be
+			 * re-issued (don't skip the instruction).
+			 */
+			pkvm_vcpu->shared_vcpu->arch.pkvm.req_param = memcache_refill_size;
+			return 0;
```
cxdong, line 15061: When we return to the host, we will sync pVM's RAX/RBX/RCX/RDX/RSI by `share_protected_vcpu_state` in vmx.c if the vmexit reason is VMCALL. But seems the host only need RAX. Maybe we should check for this hypercall in `share_protected_vcpu_state` and only share RAX?

arch/x86/kvm/mmu/mmu.c
```diff
@@ -7706,6 +7706,14 @@ void __init kvm_mmu_x86_module_init(void)
 	kvm_mmu_spte_module_init();
 }
 
+static void pkvm_mmu_vendor_module_init(void)
+{
+	if (nx_huge_pages)
+		pr_warn("disabling iTLB multihit mitigation due to pKVM enabled\n");
+	__set_nx_huge_pages(false);
+	nx_hugepage_mitigation_hard_disabled = true;
```
cxdong, line 7728: nit: It seems there is a bug bit X86_BUG_ITLB_MULTIHIT, which can be used to detect if the CPU has this bug or not. Maybe we can add this bug bit to `possible_cpu_bugs` in vmx/pkvm_init.c and always return false for it so that the pKVM will not run if the CPU has this bug? Then here it will be safe to disable the mitigation.

arch/x86/include/asm/kvm_pkvm.h
```diff
@@ -560,6 +563,12 @@ static inline size_t pkvm_guest_initial_fpstate_size(struct kvm *kvm)
 #undef KVM_BUG
 #define KVM_BUG(cond, kvm, fmt...)		KVM_BUG_ON(cond, kvm)
 
+#else /* __PKVM_HYP__ */
+
+extern bool pvmfw_present;
+extern phys_addr_t pvmfw_base;
+extern phys_addr_t pvmfw_size;
+
```
cxdong: nit: As these are defined in the arch/x86/kvm/pkvm.c from the host side, together with `enable_pkvm`, `pkvm_mem_base` and `pkvm_mem_size`, maybe we can also move these declarations to arch/x86/include/asm/kvm_host.h, which also declares for `enable_pkvm`, `pkvm_mem_base` and `pkvm_mem_size`?

arch/x86/kvm/pkvm/vmx/ept.c
```diff
@@ -398,7 +398,7 @@ int pkvm_handle_host_ept_violation(void)
 		cur.start = ALIGN_DOWN(gpa, size);
 		cur.end = cur.start + size - 1;
 
-		if (range_contains(&range, &cur)) {
+		if (range_contains(&range, &cur) && !overlaps_pvmfw(cur.start, size)) {
```
cxdong, line 401: Maybe we don't need this `overlaps_pvmfw` check, as below `pkvm_hyp_donate_host_mmio_locked(cur.start, size)` will check if `is_mmio_range(phys, size)` or not, which also performs the check `if (overlaps_pvmfw(phys, size))` by this patch?

arch/x86/include/asm/kvm_pkvm.h
```diff
@@ -498,6 +498,8 @@ static inline size_t pkvm_guest_initial_fpstate_size(struct kvm *kvm)
 	return PAGE_ALIGN(size);
 }
 
+int pkvm_vm_ioctl_enable_cap(struct kvm *kvm, struct kvm_enable_cap *cap);
```
cxdong: nit: Similar to the `pvmfw_preset`, maybe we can move this declaration to arch/x86/include/asm/kvm_host.h?

arch/x86/include/asm/kvm_pkvm.h
```diff
@@ -575,6 +577,8 @@ extern phys_addr_t pvmfw_size;
 
 static inline bool pkvm_is_protected_vm(struct kvm *kvm) { return false; }
 static inline bool pkvm_is_protected_vcpu(struct kvm_vcpu *vcpu) { return false; }
+static inline int pkvm_vm_ioctl_enable_cap(struct kvm *kvm, struct kvm_enable_cap *cap)
+{ return -EINVAL; }
```
cxdong: nit: ditto

arch/x86/include/asm/kvm_host.h
maluka-dmytro: Seems correct, but yeah, it seems more or less as easy/hard to understand as the current version. So maybe it's better to keep the current version, to minimize difference from pKVM-ARM (since the intention is that it should work exactly the same way as in pKVM-ARM)?

arch/x86/kvm/mmu/mmu.c
maluka-dmytro: Maybe... Or maybe we'd also add checking `kvm_is_mmio_pfn()` or whatever.

Anyway, I guess we'll also need to make a bunch of other changes anyway (for example, currently we always set memory type MTRR_TYPE_WRBACK), so we don't need to worry about that for now?

arch/x86/kvm/pkvm/mmu.c
maluka-dmytro: > BTW, do we have the scenario to annotate the guest mmu pte? If no, maybe can just `BUG_ON(pgt_ops->pte_annotated(ptep))`?

Yeah good point, I don't think we have such a scenario, so we can just BUG_ON (BTW like we already did in v6.12).

arch/x86/kvm/vmx/vmx.c
jaszczyk-grzegorz: Yeah that is good catch. On 6.12 it was bit different and BTW reminds me about: https://android-review.googlesource.com/c/kernel/common/+/3702540/comment/d3a45070_3987afb4/ which is a bit but not strictly related.

Here in 6.18 when things changed a bit I think we could together with skipping `WARN_ON_ONCE(kvm_skip_emulated_instruction(vcpu) != 1);` skip `kvm_rax_write(vcpu, shared_vcpu->arch.regs[VCPU_REGS_RAX]);`

If `kvm_topup_pkvm_memcache` fails or in general if `PKVM_GHC_SHARE_MEM` will fail we can't do much with current guest implementation. In 6.6. and 6.12 from which we are using kernel for guest currently there is: https://android-review.googlesource.com/c/kernel/common/+/3485332 where we always return success from SHARE/UNSHARE hypercall due to how `pkvm_set_mem_host_visibility` is written. We can still do not skip the instruction in failing case so it will be triggered again. Then if pKVM-low will notice that there is still not enough pages in the memcache it will do the same step again and again asking host to fill the memcache till there will be enough pages. Which is of course not perfect but maybe ok with current situation of deployed pVMs? WDYT?

arch/x86/kvm/pkvm/mmu.c
maluka-dmytro: Yeah indeed... The reason why I introduced these `pkvm_pgtable_lookup_range()` and `for_each_contig_range()` was to minimize host MMU walks, but somehow I forgot that here in check_guest_host_state() we don't walk host MMU at all, we only check pkvm_page->host_state which can be easily and quickly checked without extra pgtable walk. So we only need to use `pkvm_pgtable_lookup_range()` and `for_each_contig_range()` when setting host states, not when checking them.

It seems it shouldn't be too complicated to implement a custom walker for check_guest_host_state(), let me try.

maluka-dmytro: > although it introduced additional page table walk, the additional walk is not in a critial path.

Actually it is in a critical path, since this check needs to be done every time?

arch/x86/kvm/pkvm/pgtable.c
maluka-dmytro: Those are different things: "If true" refers to the `mkold` argument being true, while "if it is set" refers to the access bit in a PTE being set.

arch/x86/kvm/vmx/main.c
maluka-dmytro: Hmm, I've been thinking that we are already enforcing it indirectly, since vmx_hardware_setup() in pKVM fails with -EOPNOTSUPP if `enable_ept` (or `enable_unrestricted_guest`, `enable_vnmi` etc) is false, so pKVM will not successfully initialize in this case. But I didn't realize that pKVM does that independently of the host's settings of those params, which may be overridden by the user.

So yeah, it seems we should explicitly enforce enabling `enable_ept` in the host, for example in do_vmx_pkvm_init(), and probably do the same for `enable_unrestricted_guest`, `enable_vnmi` etc as well (at least just for consistency).

arch/x86/kvm/pkvm/mmu.c
maluka-dmytro: Yeah, indeed.

arch/x86/kvm/pkvm/mmu.c
maluka-dmytro: Right. We assume the only cases when we need to add new mappings or split existing mappings are pkvm_host_donate_guest(), pkvm_host_share_guest() and pkvm_guest_share_host(), so only in those cases we need a memcache, i.e. in other cases it can be NULL.

BTW in particular this means that we implicitly assume that the host will not request unmapping a part of an existing huge page mapping via the MMU unmap hypercall, it will only request unmapping whole mappings (so pkvm_host_unshare_guest() doesn't need to split mappings, so it doesn't need a memcache). Which in turn means, in particular, that we assume that Linux MM in the host will not trigger MMU notifiers for a part of a huge page, only for a whole huge page (since we only map a huge page for a guest if it is mapped as a huge page in the host as well).

Also pKVM-ARM has exactly the same assumptions.

arch/x86/kvm/pkvm/mmu.c
maluka-dmytro: Yeah, it seems we can.

arch/x86/kvm/pkvm/mmu.c
maluka-dmytro: > nit: maybe just `BUG_ON(for_each_contig_range(&pkvm_vm->mmu, gpa, size, __guest_unshare_host, vcpu)` and `return 0` in the end?

Yeah we can, OTOH in some other cases (in `pkvm_hyp_donate_host()` and `pkvm_host_unshare_hyp()`) we use `ret` even though we don't return it at all, we only do `BUG_ON(ret)` (and BTW I remember it was you who suggested to do that :))

> BTW, it seems the pVM guest kernel doesn't expect to receive an error code from the hypercall PKVM_GHC_UNSHARE_MEM

BTW that's a good question. In fact it doesn't even check an error for PKVM_GHC_SHARE_MEM either. And I think that's not quite fine. Since these hypercalls may return errors, it seems the guest would better check them and return to the caller (even if we assume that with a well-behaving guest those errors shouldn't happen). So I think we should clean that up when porting those guest kernel patches to v6.18.

> otherwise seems we can return the error code to the guest via the RAX rather than BUG_ON?

If unsharing fails here, it rather indicates a bug in pKVM?

arch/x86/kvm/x86.c
maluka-dmytro: I think we can, unless @jaszczyk-grzegorz (who originally implemented it this way) has any objections?

jaszczyk-grzegorz: I am fine with both. I've implemented it that way since I think it is a bit more easier to follow but if you guys think it is better in suggested way I am not against changing it - it is a bit more compact. If you decide to change it maybe it will be worth at least keeping the comment for ret = 1; saying it is handled?

arch/x86/kvm/vmx/vmx.c
maluka-dmytro: Yeah, good catch. It is still working successfully when splitting huge pages during sharing, but only because we don't write any status result to the guest RAX on the host side, so it remains with its old value i.e. the hypercall number, so here we are "overwriting" it with the value it already has.

So it works but it definitely doesn't look clean. So yeah, we'd better explicitly skip setting RAX here in this case.

arch/x86/kvm/x86.c
maluka-dmytro: Good point, we can let `share_protected_vcpu_state()` handle this case ("pkvm_hypercall_need_skip_instruction") differently.

jaszczyk-grzegorz: I think we've also discus it on 6.12 and get into conclusion that it should be fine and one of the reason it is ok is that we are zeroing other arguments during hypercalls thanks to https://android-review.googlesource.com/c/kernel/common/+/3692553 which we've also propagated to trusty. So I do not think we need to treat  hypercall differently as it should be ok. WDYT?

arch/x86/kvm/x86.c
maluka-dmytro: > maybe it will be worth at least keeping the comment for ret = 1; saying it is handled?

We don't have this comment in other such cases?

arch/x86/kvm/mmu/mmu.c
maluka-dmytro: This code is executed after KVM has already checked X86_BUG_ITLB_MULTIHIT and set `nx_huge_pages` accordingly (in kvm_mmu_x86_module_init()), so we don't need to check it again, we just use nx_huge_pages=true as an indication that the mitigation is enabled (either because of X86_BUG_ITLB_MULTIHIT or because the user has enabled it via the module param) and we are forcefully disabling it.

As for returning failure (instead of just warning, disabling the mitigation and continuing initialization), it is not that simple: it would result in failing the entire KVM initialization, not just pKVM initialization (pKVM has been already initialized at this point).

Also this CPU bug's case is quite different from those bugs which we are mitigating in `pkvm_mitigate_cpu_bugs()`: IIUC it doesn't affect security of the pKVM hypervisor, it may only allow a DoS of the host by a VM (by triggering an MCE exception causing machine reboot).

And I was thinking that it might be still useful to allow running pKVM on old CPUs affected by this bug, so it might be better to just print a warning instead of forcefully failing?

BTW here in `__set_nx_huge_pages(false)` we are also setting `itlb_multihit_kvm_mitigation` to false, which means that at least `/sys/devices/system/cpu/vulnerabilities/itlb_multihit` will correctly report that the system is vulnerable to this bug, since the mitigation is disabled.

arch/x86/include/asm/kvm_pkvm.h
maluka-dmytro: Yeah, indeed.

arch/x86/kvm/pkvm/vmx/ept.c
maluka-dmytro: If we don't check `overlaps_pvmfw` here, we will try `pkvm_hyp_donate_host_mmio_locked()` and it will fail, so we will break from the loop with a failure, without trying to map a smaller page? (So, the host is doing nothing wrong, it is accessing a MMIO address outside the pvmfw region, yet we are failing to map MMIO for the host just because we haven't even tried a mapping which is small enough to not overlap with pvmfw.)

arch/x86/kvm/x86.c
maluka-dmytro: Maybe what Chuanxiao has in mind is that this hypercall is handled mostly by pKVM, not by the host, and in those cases when it is handled also by the host, the host doesn't need to know the values of the arguments to this hypercall (RBX and RCX), it only needs to know the hypercall number (RAX).

But after thinking a bit about it, I've recalled that our approach to this is rather to handle this on a case-by-case basis for each such hypercall, not generically in `share_protected_vcpu_state()`.

In particular, when handling the PKVM_GHC_START_CPU hypercall in pKVM, before forwarding it to the host, we zero the 2nd argument (RCX) which is the CPU start address, since the host doesn't need to know it, and potentially a pVM may treat it as a secret information which should not be shared with the host.

While in this PKVM_GHC_SHARE_MEM case, both arguments (address and size of the shared region) are generally not a secret, since the host needs to know them anyway, one way or another, since it's the host who the pVM shares those pages with, to communicate with the host via virtio or whatever, thus the host obviously also needs to know which pages are those. So we don't need to do anything to hide these registers from the host?

arch/x86/kvm/vmx/main.c
maluka-dmytro: > It seems we need to enforce enable_ept in the host side, otherwise the host can still set enable_ept as 0, which will set this ops as NULL by below code in vmx_hardware_setup

BTW actually this issue is not about my patch, which adds `get_mt_mask` on the pKVM side, not on the host side? It is rather about your patch https://github.com/intel-staging/pKVM-IA/commit/aa900604c6504905a5f3e012d1ea211bb6f94ab7 ("KVM: pVMX: Implement get_mt_mask operation")

And BTW I've just realized that with pvMMU we can revert https://github.com/intel-staging/pKVM-IA/commit/aa900604c6504905a5f3e012d1ea211bb6f94ab7, right?

That said, enforcing `enable_ept`, `enable_unrestricted_guest`, `enable_vnmi`, `enable_preemption_timer` on the host side seems a good idea anyway. At least just for consistency, and maybe also to avoid various potential subtle issues caused by this inconsistency.

arch/x86/kvm/pkvm/mmu.c
maluka-dmytro: > It seems it shouldn't be too complicated to implement a custom walker for check_guest_host_state(), let me try.

I've tried, and gave up for now. :) I decided to keep it as is for now, to keep the code simpler.

arch/x86/kvm/x86.c
```diff
@@ -10653,6 +10655,26 @@ static int kvm_pkvm_hypercall(struct kvm_vcpu *vcpu)
...
+		/*
+		 * TODO: return error to the guest if refill failed due to lack
+		 * of memory. We don't do that for now, to support existing guests
+		 * which ignore any errors returned by this hypercall and thus
+		 * rely on pKVM to let the guest transparently re-issue the
+		 * hypercall instruction instead of handling -EAGAIN on the guest
+		 * side, thus we need to retain the hypercall nr in RAX and thus
+		 * can't overwrite RAX with the returned error.
+		 */
+		WARN_ON_ONCE(ret);
```
maluka-dmytro: I've just removed returning negative `ret` value here if `kvm_topup_pkvm_memcache()` failed, but then realized that it is not quite ok to remove it: if we are returning this error, it is returned to userspace so the VMM can handle it (most likely just by killing the VM), which is probably not as good as returning this error to the guest, but still better than not returning it at all. Let me fix this.

arch/x86/kvm/vmx/vmx.c
maluka-dmytro: Updated, hopefully now the code looks cleaner.

> BTW, what if the `kvm_topup_pkvm_memcache` is failed?

I think the error will be returned to userspace, and then the VMM most likely will kill the VM. So it is probably not as good as returning the error to the guest itself, but probably good enough for now (and it allows, again, to avoid modifying our existing pVMs).

arch/x86/include/asm/kvm_host.h
cxdong: Yeah, makes sense to minimize difference from pKVM-ARM.

arch/x86/kvm/mmu/mmu.c
cxdong: Yeah, we can consider that when passthrough is needed.

arch/x86/kvm/pkvm/mmu.c
cxdong: > Actually it is in a critical path, since this check needs to be done every time?

I see `check_guest_host_state` is called by `pkvm_host_unshare_guest`, `pkvm_guest_share_host` and `pkvm_guest_unshare_host`. The first one is used when destroy a npVM which doesn't happen frequently. The other two are used when a pVM explicitly sharing its memory with the host, and this also doesn't happen frequently due to using bounce buffer. So I think it is not in a critical path.

> I've tried, and gave up for now. :) I decided to keep it as is for now, to keep the code simpler.

Yeah, agree.

arch/x86/kvm/pkvm/pgtable.c
cxdong: Ah, I see. I misunderstood this.

arch/x86/kvm/vmx/main.c
cxdong: > BTW actually this issue is not about my patch, which adds get_mt_mask on the pKVM side, not on the host side? It is rather about your patch https://github.com/intel-staging/pKVM-IA/commit/aa900604c6504905a5f3e012d1ea211bb6f94ab7 ("KVM: pVMX: Implement get_mt_mask operation")

Yes, it is about this patch. 

> And BTW I've just realized that with pvMMU we can revert https://github.com/intel-staging/pKVM-IA/commit/aa900604c6504905a5f3e012d1ea211bb6f94ab7, right?

Indeed, so we don't need to use that after enabling the pvMMU. This patch can be reverted as part of this serial?

> That said, enforcing enable_ept, enable_unrestricted_guest, enable_vnmi, enable_preemption_timer on the host side seems a good idea anyway. At least just for consistency, and maybe also to avoid various potential subtle issues caused by this inconsistency.

`enable_vnmi` is not used by the host actually. And the `enable_preemption_timer` can be either true or false from the host side. If it is true, the host will use preemption timer to emulate guest timer, which is just supported by the pvvmcs part 3. And if it is false, the host can still use software timer to emulate guest timer, which is also supported. 
The `enable_ept` and `enable_unrestricted_guest` are used by the host.

So maybe we just need to enforced `enable_ept` and `enable_unrestricted_guest` to align with the pKVM?

arch/x86/kvm/pkvm/mmu.c
cxdong: Thanks for the explaination!

cxdong: BTW, for the `__guest_share_host`, it may allocate memory pages from `&vcpu->arch.pkvm.guest_mmu_memcache`. How do we guarantee `&vcpu->arch.pkvm.guest_mmu_memcache` have enough memory pages for this hypercall from the guest?

arch/x86/kvm/pkvm/mmu.c
cxdong: > Yeah we can, OTOH in some other cases (in pkvm_hyp_donate_host() and pkvm_host_unshare_hyp()) we use ret even though we don't return it at all, we only do BUG_ON(ret) (and BTW I remember it was you who suggested to do that :))

That was for the case that, we need to `BUG_ON()` for quite some function calls in one function, to avoid using many `BUG_ON()`, just use one single `BUG_ON(ret)` in the end of the function before return.

Here we just have one single `BUG_ON` so I think it can be simplified as `BUG_ON(for_each_contig_range(&pkvm_vm->mmu, gpa, size, __guest_unshare_host, vcpu)`. Either way is fine to me, it is just a coding style.

> So I think we should clean that up when porting those guest kernel patches to v6.18.

Yeah, agree.

> If unsharing fails here, it rather indicates a bug in pKVM?

Yes, if it is unsharing, we can just BUG_ON. The other error code, e.g., alignment checking, page state checking, should be handled by the guest.

BTW, we have an error recovery code path in the `pkvm_guest_share_host` to unshare the already shared memory pages. Do we also want to use a BUG_ON for that `__guest_unshare_host`?
```
          ret = for_each_contig_range(&pkvm_vm->mmu, gpa, size,
                                      __guest_share_host, vcpu);
          if (ret) {
                  /* Unshare already shared pages, if any. */
                  for_each_contig_range(&pkvm_vm->mmu, gpa, size,
                                        __guest_unshare_host, vcpu);
          }
```

arch/x86/kvm/x86.c
cxdong: > I think we've also discus it on 6.12 and get into conclusion that it should be fine and one of the reason it is ok is that we are zeroing other arguments during hypercalls thanks to https://android-review.googlesource.com/c/kernel/common/+/3692553 which we've also propagated to trusty. 

This patch is to let the guest itself to guarantee it doesn't accidently send any secret via its registers to the host. For example, the guest sends a hypercall to the host with only 2 arguments in RBX/RCX, so the RDX/RSI are not used. To simplify, the pKVM hypervisor assume a hypercal can mostly have 4 arguments, so the pKVM will share RBX/RCX/RDX/RSI with the host. So if the guest sends a hypercall which only has 2 arguments, it should zero the other 2 registers as well.

And `share_protected_vcpu_state()` was implemented for the hypercalls which are not handled by the pKVM but handled by the host. Now we have a special usage that, some hypercall may be handled by the pKVM first, and if the pKVM cannot handle that due to some reason and need to involve the host, we forward that hypercall to the host, still via `share_protected_vcpu_state()`. And it will share RAX/RBX/RCX/RDX/RSI to the host. In this case, we will need to pay attension on the shared registers to see if they contain any secret, which should not be shared witht he host.

>But after thinking a bit about it, I've recalled that our approach to this is rather to handle this on a case-by-case basis for each such hypercall, not generically in share_protected_vcpu_state().

Yeah, if we can deal with the registers before going to `share_protected_vcpu_state()`, then we don't need to do that again in `share_protected_vcpu_state()`.

> While in this PKVM_GHC_SHARE_MEM case, both arguments (address and size of the shared region) are generally not a secret, since the host needs to know them anyway, one way or another, since it's the host who the pVM shares those pages with, to communicate with the host via virtio or whatever, thus the host obviously also needs to know which pages are those. So we don't need to do anything to hide these registers from the host?

Indeed the address and size of the shared region are not a secret. Then we don't need to do anything here.

arch/x86/kvm/mmu/mmu.c
cxdong: > Also this CPU bug's case is quite different from those bugs which we are mitigating in pkvm_mitigate_cpu_bugs(): IIUC it doesn't affect security of the pKVM hypervisor, it may only allow a DoS of the host by a VM (by triggering an MCE exception causing machine reboot).

I see. Thanks for pointing this out. So we are fine just printing a warning and disable the mitigation if the CPU has this bug, or the mitigation is force enabled.

arch/x86/kvm/pkvm/vmx/ept.c
cxdong: I see. My understanding was that, the pvmfw range is a hole we punch in the memory, so the range find out by `pkvm_find_addr_range(gpa, &range) || is_pvmfw(gpa)` will not be overlapped with pvmfw, and `overlaps_pvmfw(cur.start, size)` is a safety check. But seems it is possible that the pvmfw can just located at the edge of a memory arae, or in extreme case, the pvmfw is not besides any memory area, so we still need `overlaps_pvmfw(cur.start, size)` to filter out the pvmfw range before mapping for MMIO, right?

arch/x86/kvm/pkvm/mmu.c
maluka-dmytro: > The first one is used when destroy a npVM which doesn't happen frequently.

Not only, it is also triggered quite often under low-memory conditions, when the host is migrating or swapping a lot of npVM pages. Or for example when inflating virtio-balloon, and BTW with crosvm's flag `--balloon-page-reporting` that happens all the time during normal operation, not just under low-memory conditions (although I didn't measure how frequently it happens then).

BTW side note: on the host side we don't have it quite optimized either: in `kvm_unmap_gfn_range()` we call the unmap hypercall to pKVM for each mapping in the range, so if the range is large, it takes a lot of time. Furthermore, that is done with kvm->mmu_lock spinlock write-held. So, for instance, it is causing failure of the KVM selftest `mmu_stress_test` (mprotect() syscall triggers MMU notifier for a 128GB gpa range, i.e. 32 million 4K mappings => `kvm_unmap_gfn_range()` makes 32 million hypercalls => exceeding test timeout of 120 seconds), plus causing CPU stalls on nearly all CPUs for many seconds during the test (since on one CPU kvm->mmu_lock is locked by `kvm_unmap_gfn_range()` which is busy making all those hypercalls, and on other CPUs `pkvm_page_fault()` is trying to handle page faults of the test VM's vCPUs and is waiting on kvm->mmu_lock).

And AFAICS pKVM-ARM should have exactly the same problem, since it implements that exactly the same way. So I assume we can live with it for now. :)

> The other two are used when a pVM explicitly sharing its memory with the host, and this also doesn't happen frequently due to using bounce buffer.

Currently yes, but potentially it could be used without bounce buffer as well, i.e. zero-copy, for some specific performance-critical use cases (e.g. DRM playback or whatever)?

cxdong: > BTW side note: on the host side we don't have it quite optimized either: in kvm_unmap_gfn_range() we call the unmap hypercall to pKVM for each mapping in the range, so if the range is large, it takes a lot of time. Furthermore, that is done with kvm->mmu_lock spinlock write-held. So, for instance, it is causing failure of the KVM selftest mmu_stress_test (mprotect() syscall triggers MMU notifier for a 128GB gpa range, i.e. 32 million 4K mappings => kvm_unmap_gfn_range() makes 32 million hypercalls => exceeding test timeout of 120 seconds), plus causing CPU stalls on nearly all CPUs for many seconds during the test (since on one CPU kvm->mmu_lock is locked by kvm_unmap_gfn_range() which is busy making all those hypercalls, and on other CPUs pkvm_page_fault() is trying to handle page faults of the test VM's vCPUs and is waiting on kvm->mmu_lock).

I see. Thanks for the notice!

>Currently yes, but potentially it could be used without bounce buffer as well, i.e. zero-copy, for some specific performance-critical use cases (e.g. DRM playback or whatever)?

But suppose they should also not frequently send the SHARE/UNSHARE hypercalls as doing so will also have performance impact.

arch/x86/kvm/vmx/vmx.c
cxdong: Yeah, it looks good to me.

arch/x86/kvm/vmx/main.c
maluka-dmytro: > Indeed, so we don't need to use that after enabling the pvMMU. This patch can be reverted as part of this serial?

Yes, I've already done that in https://github.com/intel-staging/pKVM-IA/pull/66/commits/f852635be488ad2aea99cd1dadc8183dd998b5c8 in the last version of this PR.

> `enable_vnmi` is not used by the host actually.

But it doesn't hurt to force it anyway? In case it is somehow used by the host in the future, or e.g. just to see an accurate value in `cat /sys/module/kvm_intel/parameters/vnmi`.

> And the enable_preemption_timer can be either true or false from the host side. [...] And if it is false, the host can still use software timer to emulate guest timer, which is also supported.

Ok, I've been thinking that if it is false in the host, the behavior is inconsistent (so I've already added forcing it to true in https://github.com/intel-staging/pKVM-IA/pull/66/commits/e6922eb77ef39d9d9050766e94f9b36b2cf90c07). Looks like I've been wrong, so we shouldn't force it.

arch/x86/kvm/pkvm/mmu.c
cxdong: > BTW, we have an error recovery code path in the pkvm_guest_share_host to unshare the already shared memory pages. Do we also want to use a BUG_ON for that __guest_unshare_host?

I see the latest code has already done this.

arch/x86/kvm/pkvm/mmu.c
maluka-dmytro: The patch https://github.com/intel-staging/pKVM-IA/pull/66/commits/93daa2129659f2243d0289d81bbf2dc32b7c3eac ("pKVM: x86: handle memcache refill during page sharing"), which we've been discussing in detail, is all about guaranteeing that. :)

As explained in comments in that patch, before calling `pkvm_guest_share_host()`, pKVM first checks if there are enough pages in the memcache. If not, it forwards the guest's hypercall to the host to let it refill the memcache, without even trying to call `pkvm_guest_share_host()`. After that the guest issues the hypercall again (and this time there are enough pages, otherwise pKVM will ask the host to refill the memcache once again).

arch/x86/kvm/vmx/main.c
cxdong: > Yes, I've already done that in https://github.com/intel-staging/pKVM-IA/commit/f852635be488ad2aea99cd1dadc8183dd998b5c8 in the last version of this PR.

Got it. Thanks.

>But it doesn't hurt to force it anyway? In case it is somehow used by the host in the future, or e.g. just to see an accurate value in cat /sys/module/kvm_intel/parameters/vnmi.

Yes, it doesn't hurt, and makes sense. Let's keep it enforced as well.

arch/x86/kvm/pkvm/mmu.c
maluka-dmytro: > BTW, we have an error recovery code path in the `pkvm_guest_share_host` to unshare the already shared memory pages. Do we also want to use a BUG_ON for that `__guest_unshare_host`?

Yes, you already suggested that in another comment, and I've already done that in https://github.com/intel-staging/pKVM-IA/pull/66/commits/499584de052709f638cdcc4e1988e5f7cc7a73c9 as you suggested.

arch/x86/kvm/pkvm/mmu.c
cxdong: I see. Sorry, just forget about the part of handling the hypercall before sharing the guest. :)

arch/x86/kvm/pkvm/vmx/ept.c
maluka-dmytro: Yes, pvmfw can be anywhere. In particular, pvmfw is a reserved memory region (bootloader reserves it in e820), so from pKVM perspective it is not "memory", and I imagine it is not unlikely that pvmfw may be, for example, between two other reserved memory regions (while pKVM treats both those reserved memory regions just like MMIO, in particular allows the host to lazily map them).

arch/x86/kvm/vmx/main.c
maluka-dmytro: Updated to not force `enable_preemption_timer`.

arch/x86/kvm/pkvm/mmu.c
maluka-dmytro: > But suppose they should also not frequently send the SHARE/UNSHARE hypercalls as doing so will also have performance impact.

Yeah, OTOH maybe there may be a tradeoff such that the frequency of those hypercalls is "medium": low enough so that there is a performance win due to zero-copy, but high enough so that the cost of the hypercall is still not negligible...

arch/x86/kvm/pkvm/mmu.c
```diff
@@ -1266,6 +1292,26 @@ int pkvm_host_donate_guest(struct kvm_vcpu *vcpu, unsigned long gpa,
 
 	set_host_mem_pgstate(hpa, size, PKVM_PAGE_NONE);
 
+	if (gpa_range_overlaps_pvmfw(&pkvm_vm->kvm, gpa, size, &gpa_offset,
+				     &pvmfw_offset, &load_size)) {
+		/*
+		 * Make sure pvmfw is loaded into the guest memory pages after
+		 * enabling protection of these pages from the host, not before.
+		 */
+		smp_wmb();
```
cxdong: nit: Does the `after enabling protection of these pages from the host` mean for the protection done by `pkvm_pgtable_set_owner(&host_mmu, hpa, size, PKVM_ID_GUEST)` and `set_host_mem_pgstate(hpa, size, PKVM_PAGE_NONE);`? If so, any reason to prefer a `smp_wmb()` rather than a `wmb()`?

cxdong (approved):
This serial LGTM. Thanks!

PR #68, opened 2026-02-03
Pkvm v6.18 fix warnings

cxdong:
Fix the compiling warnings of unused-but-set-variable in the vmx.c for the pKVM. This may not be that important to be fixed as the other components in the linux kernel also have such warnings as below, so the linux kernel still cannot pass the compiling even we fixed for the pKVM. But maybe it is still better to do that regardless this, although it introduced a bit more `#ifndef __PKVM_HYP__` in the code.

```
../drivers/acpi/acpica/nsaccess.c:295:6: error: variable 'num_carats' set but not used [-Werror,-Wunused-but-set-variable]
  295 |         u32 num_carats;
      |             ^
1 error generated.
make[6]: *** [../scripts/Makefile.build:287: drivers/acpi/acpica/nsaccess.o] Error 1
make[6]: *** Waiting for unfinished jobs....
../arch/x86/kernel/process.c:972:10: error: variable 'hi' set but not used [-Werror,-Wunused-but-set-variable]
  972 |         u32 lo, hi;
      |                 ^
1 error generated.       ^
```

### DIFF
--- a/arch/x86/kvm/vmx/vmx.c
+++ b/arch/x86/kvm/vmx/vmx.c
@@ -6235,10 +6235,14 @@ static int handle_task_switch(struct kvm_vcpu *vcpu)
 {
 	struct vcpu_vmx *vmx = to_vmx(vcpu);
 	unsigned long exit_qualification;
+#ifndef __PKVM_HYP__
 	bool has_error_code = false;
 	u32 error_code = 0;
 	u16 tss_selector;
 	int reason, type, idt_v, idt_index;
+#else
+	int reason, type, idt_v;
+#endif
 
 #ifdef __PKVM_HYP__
 	/*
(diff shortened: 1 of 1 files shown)


maluka-dmytro:
How about just something like the following instead (to avoid all those extra ugly #ifdefs):
```
--- a/arch/x86/kvm/vmx/vmx.c
+++ b/arch/x86/kvm/vmx/vmx.c
@@ -5719,6 +5719,8 @@ static int handle_exception_nmi(struct kvm_vcpu *vcpu)
                } else
                        return kvm_handle_page_fault(vcpu, error_code, cr2, NULL, 0);
        }
+#else
+       (void)vect_info;
 #endif
 
        ex_no = intr_info & INTR_INFO_VECTOR_MASK;
@@ -6304,6 +6306,11 @@ static int handle_task_switch(struct kvm_vcpu *vcpu)
                               type == INTR_TYPE_SOFT_INTR ? idt_index : -1,
                               reason, has_error_code, error_code);
 #else
+       (void)has_error_code;
+       (void)error_code;
+       (void)tss_selector;
+       (void)idt_index;
+
        /*
         * Switch to the host and let the host to emulate the task switch
         * for the npVMs.
```

But actually I'm thinking: so apparently Linux kernel devs don't care about compiling with `-Wunused-but-set-variable`, so perhaps we shouldn't care either, and prefer simpler code?

jaszczyk-grzegorz:
I also think that it is better to not merge this one to not get more ifdefs in the common code as Linux kernel doesn't care about `-Wunused-but-set-variable`. The merge conflict surface will be smaller

cxdong:
Sure, let's close this PR until we need it.

PR #69, opened 2026-02-04
SQUASHME: pKVM: x86: Use pkvm_is_protected_vm() before vcpu is initia…

Qiangcy:
…lized.

Since vcpu->kvm is initialized after this check,
pkvm_is_protected_vcpu() is not functional at this stage.

Fixes: 6fdd05cd1022 ("pKVM: VMX: Synchronize host disabled_exits to pKVM in npVMs")

### DIFF
--- a/arch/x86/kvm/pkvm/pkvm.c
+++ b/arch/x86/kvm/pkvm/pkvm.c
@@ -391,7 +391,7 @@ static int __vcpu_create(struct kvm *kvm, struct kvm_vcpu *vcpu, struct fpstate 
 		kvm->arch.notify_window = pkvm_vm->shared_kvm->arch.notify_window;
 		kvm->arch.notify_vmexit_flags = pkvm_vm->shared_kvm->arch.notify_vmexit_flags;
 	}
-	if (!pkvm_is_protected_vcpu(vcpu))
+	if (!pkvm_is_protected_vm(kvm))
 		kvm->arch.disabled_exits = pkvm_vm->shared_kvm->arch.disabled_exits;
 
 	pkvm_spin_unlock(&pkvm_vm->lock);


cxdong (approved):
It seems that, without setting vcpu->kvm, `pkvm_is_protected_vcpu(vcpu)` doesn't trigger any page fault to stop the pKVM, is because the vcpu->kvm.arch.vm_type may be located at some virtual address in the vmemmap which is mapped in the hypervisor's MMU.

The fix LGTM.

maluka-dmytro (approved):

maluka-dmytro:
Good catch, thanks.

Please note that we've just merged pvVMCS-part3 to the Android 6.18 branch yesterday (https://android-review.googlesource.com/q/topic:%22pKVM-IA-pvVMCS-3%22), so I assume we should just merge this patch, without squashing it.

maluka-dmytro:
> It seems that, without setting vcpu->kvm, `pkvm_is_protected_vcpu(vcpu)` doesn't trigger any page fault to stop the pKVM, is because the vcpu->kvm.arch.vm_type may be located at some virtual address in the vmemmap which is mapped in the hypervisor's MMU.

Yeah, out of curiosity I've just printed out the offset of the `arch` field in `struct kvm` and it is 5512, so even though the `vcpu->kvm` pointer is 0, the address of `vcpu->kvm->arch.vm_type` is mapped (as vmemmap), since it is in page 1, not in page 0.

cxdong:
> Good catch, thanks.
> 
> Please note that we've just merged pvVMCS-part3 to the Android 6.18 branch yesterday (https://android-review.googlesource.com/q/topic:%22pKVM-IA-pvVMCS-3%22), so I assume we should just merge this patch, without squashing it.

Yes, I will merge directly without squashing it.

PR #71, opened 2026-02-11
SQUASHME: KVM: x86/mmu: pKVM: Don't skip kvm_arch_flush_shadow_all()

maluka-dmytro:
Currently host KVM skips zapping all guest mappings in kvm_arch_flush_shadow_all() during VM teardown is pKVM is enabled, assuming that it is not needed since pKVM hypervisor is about to destroy the guest MMU anyway, when handling the vm_destroy hypercall.

However, the host kernel may start using guest pages for other needs right after kvm_arch_flush_shadow_all(), before vm_destroy, since KVM executes kvm_arch_flush_shadow_all() as a part of the MMU notifier release, thus telling the host kernel's MM that the pages are available to use for other needs.

This is not an issue for pVMs, since pVM pages are still pinned (they are only unpinned after vm_destroy). However, this is an issue for npVMs, since npVM pages state for the host is still SHARED_OWNED, not OWNED, thus the host will fail e.g. to donate those pages to a pVM if it tries to do that immediately after kvm_arch_flush_shadow_all().

So don't skip kvm_arch_flush_shadow_all() but let it properly ask pKVM to unmap guest pages and thus properly update their host state to OWNED, so that the host kernel can immediately start using them for other needs.

Fixes: 30e9a30c5f52 ("KVM: x86/mmu: pKVM: Skip kvm_arch_flush_shadow_all()")

### DIFF
--- a/arch/x86/kvm/mmu/mmu.c
+++ b/arch/x86/kvm/mmu/mmu.c
@@ -1645,7 +1645,7 @@ static bool __kvm_rmap_zap_gfn_range(struct kvm *kvm,
 				 start, end - 1, can_yield, true, flush);
 }
 #ifdef CONFIG_PKVM_X86
-static bool pkvm_unmap_gfn_range(struct kvm *kvm, struct kvm_gfn_range *range)
+static bool __pkvm_unmap_gfn_range(struct kvm *kvm, gfn_t start, gfn_t end)
 {
 	struct pkvm_mapping *m;
 
@@ -1654,7 +1654,7 @@ static bool pkvm_unmap_gfn_range(struct kvm *kvm, struct kvm_gfn_range *range)
 	if (pkvm_is_protected_vm(kvm))
 		return false;
 
-	for_each_pkvm_mapping(kvm, range->start, range->end, m) {
+	for_each_pkvm_mapping(kvm, start, end, m) {
 		int err = pkvm_hypercall(vm_mmu_unmap, kvm->arch.pkvm.handle,
 					 m->gfn << PAGE_SHIFT,
 					 m->nr_pages << PAGE_SHIFT);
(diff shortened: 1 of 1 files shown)


maluka-dmytro:
As usual, SQUASHME: is just for future rebases to newer kernels, not for squashing it now.

cxdong (approved):
Good cache. This issue seems not easy to be reproduced if we don't stress booting npVMs and pVMs.

maluka-dmytro:
In fact it was reproduced by stress testing pvIOMMU. Namely, by killing a npVM while a DMA map/unmap stress test is running in the background. In current WIP pvIOMMU version, before allowing to map pages for DMA, pKVM checks if the pages host state is PKVM_PAGE_OWNED. In this corner case the state is PKVM_PAGE_SHARED_OWNED instead, so DMA map fails.

I'm actually thinking why shouldn't we allow DMA map for PKVM_PAGE_SHARED_OWNED (i.e. for npVM pages) as well. Anyway, not allowing it has helped us to catch this issue.

maluka-dmytro:
Minor thing: @jaszczyk-grzegorz noticed a typo in the commit message:

s/is pKVM is enabled/if pKVM is enabled/

@cxdong maybe you could force-push it with this correction? (Sorry for the trouble :))

cxdong:
@maluka-dmytro, I will do a force update after we merge the recent opened PRs, to avoid updating them.

maluka-dmytro:
Sure, thanks.

BTW while we're at it, another minor thing, regarding the TSC frequency patch (https://github.com/intel-staging/pKVM-IA/pull/72): before we merged it on gerrit, Greg also noticed that "TSC crystal clock" in the comments sounds inaccurate/misleading, so I changed it to "core crystal clock": https://android-review.googlesource.com/c/kernel/common/+/3956675/1..3/arch/x86/kvm/cpuid.c, so we could force-update that as well?

cxdong:
I will do that as well.

cxdong:
@maluka-dmytro , pkvm-v6.18 is force updated with these two minor issues fixed. Thanks!

PR #72, opened 2026-02-16
pKVM: x86: Expose TSC frequency to pVM

maluka-dmytro:
Since pKVM does not support kvmclock for protected VMs, a pVM can't find out the TSC frequency from the host directly. So Linux pVMs find it out by calibrating the TSC against a PIT or HPET timer emulated by the host, which is inherently unreliable: if the host is under a heavy load during this calibration, the calibration will produce inaccurate results or, more likely, will just fail. And if the calibration fails, the guest will deem the TSC unstable (whereas actually it is the timer that is inaccurate, while the TSC is accurate, since it is native), and will fall back to using a jiffies-based clocksource instead of TSC, thus requiring periodic timer interrupts and thus causing permanently high CPU usage as long as the pVM is running.

Furthermore, on PTL (not on older CPUs) for some reason this problem (failed TSC calibration) is observed all the time, even if the system is not under a high load when the pVM starts.

So to avoid this problem and yet avoid special changes on the guest side, enforce exposing the TSC frequency info to the pVM via the CPUID leaf 0x15, so that the pVM doesn't need to calibrate the TSC.

In the future this enforcement might be also reused for security purposes, as a part of providing secure TSC for pVMs (which pKVM doesn't provide yet).

### DIFF
--- a/arch/x86/kvm/cpuid.c
+++ b/arch/x86/kvm/cpuid.c
@@ -2138,6 +2138,38 @@ bool kvm_cpuid(struct kvm_vcpu *vcpu, u32 *eax, u32 *ebx,
 				*eax = vcpu->arch.hw_tsc_khz;
 			}
 #endif
+		} else if (pkvm_is_protected_vcpu(vcpu) && function == 0x15) {
+			/*
+			 * Since protected VMs cannot use kvmclock to find out the
+			 * TSC frequency, expose the native TSC frequency directly
+			 * via cpuid. For now this is not for security (pKVM does not
+			 * provide secure TSC yet) but just to let the guest know the
+			 * TSC frequency so that it doesn't need to calibrate the TSC
+			 * against host-emulated PIT/HPET timers which are inherently
+			 * inaccurate as they are sensitive to the system load in the
+			 * host.
+			 */
+			if (tsc_khz) {
+				/*
+				 * TSC crystal clock must match the LAPIC clock.
+				 * See native_calibrate_tsc().
+				 */
+				*ecx = 1000000000 / vcpu->kvm->arch.apic_bus_cycle_ns;
+				/*
+				 * Try to minimize calculation error: use 38.4 MHz
+				 * which is the typical TSC crystal clock frequency
+				 * on today's Intel's desktop CPUs.
+				 *
+				 * TODO: we could avoid any calculation error, by
+				 * exposing tsc_khz via some PV (akin to kvmclock)
+				 * instead of cpuid.
+				 */
+				*eax = *ecx / 38400000;	/* denominator */
+				*ebx = tsc_khz / 38400;	/* numerator */
+				*edx = 0;
+			} else {
+				*eax = *ebx = *ecx = *edx = 0;
+			}
 		}
 	} else {
 		*eax = *ebx = *ecx = *edx = 0;
--- a/arch/x86/kvm/pkvm/pkvm.c
+++ b/arch/x86/kvm/pkvm/pkvm.c
@@ -163,6 +163,7 @@ static int pkvm_vm_init(phys_addr_t host_kvm_pa, phys_addr_t pkvm_vm_pa,
 		kvm->arch.disabled_quirks = (kvm_caps.inapplicable_quirks |
 					     pkvm_vm->shared_kvm->arch.disabled_quirks) &
 					    kvm_caps.supported_quirks;
+	kvm->arch.apic_bus_cycle_ns = APIC_BUS_CYCLE_NS_DEFAULT;
 	kvm->arch.pkvm.pvmfw_load_addr = INVALID_GPA;
 
 	pkvm_spin_lock_init(&pkvm_vm->lock);
@@ -462,6 +463,8 @@ static int __vcpu_create(struct kvm *kvm, struct kvm_vcpu *vcpu, struct fpstate 
 		kvm->arch.notify_window = pkvm_vm->shared_kvm->arch.notify_window;
 		kvm->arch.notify_vmexit_flags = pkvm_vm->shared_kvm->arch.notify_vmexit_flags;
 	}
+	if (pkvm_vm->shared_kvm->arch.apic_bus_cycle_ns)
+		kvm->arch.apic_bus_cycle_ns = pkvm_vm->shared_kvm->arch.apic_bus_cycle_ns;
 	if (!pkvm_is_protected_vm(kvm))
 		kvm->arch.disabled_exits = pkvm_vm->shared_kvm->arch.disabled_exits;
 


arch/x86/kvm/cpuid.c
```diff
@@ -2138,6 +2138,38 @@ bool kvm_cpuid(struct kvm_vcpu *vcpu, u32 *eax, u32 *ebx,
 				*eax = vcpu->arch.hw_tsc_khz;
 			}
 #endif
+		} else if (pkvm_is_protected_vcpu(vcpu) && function == 0x15) {
```
cxdong, line 2141: If we disable pv features for the npVM, it seems the npVM also has the same issue. Do we also consider for this case?

> Furthermore, on PTL (not on older CPUs) for some reason this problem (failed TSC calibration) is observed all the time, even if the system is not under a high load when the pVM starts.

It seems this is something worth to understand the root cause. I can take a look at this at the next week after back from the holiday.

maluka-dmytro: > If we disable pv features for the npVM, it seems the npVM also has the same issue.

Yeah, but then it has the same issue without pKVM as well, so it is not our concern? :)

As I understand it, this underlying issue is inevitable if the VM has no way to know the TSC frequency reliably and thus has to "guess" it. So the solution is to reliably provide the TSC frequency info to the VM, one way or another. For a npVM, kvmclock provides it.

BTW while implementing this patch I realized that the approach of this patch (i.e. using CPUID) is not perfect, since it implies some calculation error due to division (as mentioned in the comment here in the code). In my tests on several different CPUs the error is 4-5MHz (e.g. 2692 MHz instead of 2688 MHz), i.e. just 0.1-0.2%, so it shouldn't be a big problem for our current use cases. But still, it is entirely possible to provide 100% accurate frequency, without any error, so it would be probably better to do that. But that would probably require changes in the guest.

So I've been thinking: if we ever support secure TSC, we'd probably provide TSC frequency from pKVM to pVM via some hypercall (and BTW also, for example, this hypercall returning non-zero frequency could indicate to the pVM that secure TSC is available).

And in the meantime while we don't provide secure TSC, I guess we could just add kvmclock support for pVM. Perhaps that would be better than this patch (it would provide precise frequency without calculation error, plus we still wouldn't need to define our own non-standard PV interface for that). And BTW, I'm not really familiar with kvmclock, but from the code in `arch/x86/kernel/kvmclock.c` it seems it might be easy to implement it: it seems we'd just need to let the guest share its `.bss..decrypted` section with the host (so that the host can access the guest's `hv_clock_boot`)?

However, still, the "advantage" of this patch is that it allows to avoid any guest changes.

maluka-dmytro: > It seems this is something worth to understand the root cause.

Yeah, agree.

That said, even if this wasn't observed on PTL (i.e. even if the issue occurs only sporadically under a high load), I think this would be still a rather critical issue: if we have a long-running pVM (e.g. running all the time in the background), and if we are unlucky and the issue happens when this pVM starts (e.g. during system boot), this pVM is gonna consume a lot of CPU all the time, thus still causing serious performance & power degradation.

cxdong: > Yeah, but then it has the same issue without pKVM as well, so it is not our concern? :)

With disabling pKVM, the npVM without pv feature seems doesn't have this issue on PTL.

>And in the meantime while we don't provide secure TSC, I guess we could just add kvmclock support for pVM.

The kvmclock driver in the guest relies on KVM signature, which is overridened to PKVM signature. If we want to reuse any KVM PV feature for a pVM, it is better we use an alternative to let the guest kernel to identify the pVM rather than the PKVM signature.

maluka-dmytro: > With disabling pKVM, the npVM without pv feature seems doesn't have this issue on PTL.

Hmm, interesting.

BTW maybe it is also related to another observation: I remember that with the old pre-pvVMCS implementation of pKVM-IA (i.e. based on VMX emulation and shadow VMCS), when I was running pKVM in qemu, I didn't see any problems with continuous high CPU consumption by pVMs. And then after switching to pvVMCS, I started seeing this problem in qemu. (Back then I didn't know what was the cause, I just saw high CPU usage and a lot of hypercalls from host to pKVM. Later I found out that those hypercalls were for interrupt injection, and the cause was the pVM using periodic timer.)

> The kvmclock driver in the guest relies on KVM signature, which is overridened to PKVM signature. If we want to reuse any KVM PV feature for a pVM, it is better we use an alternative to let the guest kernel to identify the pVM rather than the PKVM signature.

Yeah, so perhaps we could add kvmclock support as a part of this "cleanup" (switching from PKVMPKVMPKVM to reusing KVMKVMKVM) in the future.

cxdong: >That said, even if this wasn't observed on PTL (i.e. even if the issue occurs only sporadically under a high load), I think this would be still a rather critical issue: if we have a long-running pVM (e.g. running all the time in the background), and if we are unlucky and the issue happens when this pVM starts (e.g. during system boot), this pVM is gonna consume a lot of CPU all the time, thus still causing serious performance & power degradation.

Makes sense. It is worth to fix this by this patch for the pVM. Still, maybe also can do the same for the npVM in case the npVM doesn't enable the kvmclock feature? Perhaps this is not important for us as we always boot the npVM with the kvmclock feature enabled on PTL?

> BTW maybe it is also related to another observation: I remember that with the old pre-pvVMCS implementation of pKVM-IA (i.e. based on VMX emulation and shadow VMCS), when I was running pKVM in qemu, I didn't see any problems with continuous high CPU consumption by pVMs. And then after switching to pvVMCS, I started seeing this problem in qemu. (Back then I didn't know what was the cause, I just saw high CPU usage and a lot of hypercalls from host to pKVM. Later I found out that those hypercalls were for interrupt injection, and the cause was the pVM using periodic timer.)

My first impression about this is that, injecting interrupt with the shadow-VMCS implementation may have less number of vmexits comparing with the pv-VMCS, as writing VMCS field with vmwrite instruction doesn't have vmexit with shadow-VMCS, while we have to use a hypercall with pv-VMCS.

>Yeah, so perhaps we could add kvmclock support as a part of this "cleanup" (switching from PKVMPKVMPKVM to reusing KVMKVMKVM) in the future.

Agree.

maluka-dmytro: > Perhaps this is not important for us as we always boot the npVM with the kvmclock feature enabled on PTL?

I think so. We can apply this fix to npVM as well (by removing this `pkvm_is_protected_vcpu(vcpu)` check) if we want to, but it seems there is not much practical need for that. (And per the above, this fix might be actually just a temporary solution anyway.)

> My first impression about this is that, injecting interrupt with the shadow-VMCS implementation may have less number of vmexits comparing with the pv-VMCS, as writing VMCS field with vmwrite instruction doesn't have vmexit with shadow-VMCS, while we have to use a hypercall with pv-VMCS.

Sounds plausible. BTW I recall we discussed we could try to optimize this by "batching" interrupt injection requests into the vcpu_run hypercall, to avoid extra hypercalls for them...

Now, if this is indeed the cause of the problem in case of qemu, the question is if it is the cause in case of PTL as well. I guess we could try to confirm that by running pre-pvVMCS implementation on PTL.

And if this is indeed the cause on PTL as well, it is not clear why PTL is more prone to this than older CPUs...

cxdong: > I think so. We can apply this fix to npVM as well (by removing this pkvm_is_protected_vcpu(vcpu) check) if we want to, but it seems there is not much practical need for that. (And per the above, this fix might be actually just a temporary solution anyway.)

OK. Then we can just apply for pVM.

cxdong (approved):

PR #73, opened 2026-02-17
Pkvm v6.18 fix linking after lts merge

jaszczyk-grzegorz:
After LTS patch "KVM: x86: Explicitly configure supported XSS from {svm,vmx}_set_cpu_caps()" was merged into 6.18 the pKVM failed to build. To PR is to fix regression. It was also already merged in: https://android-review.googlesource.com/c/kernel/common/+/3951504


cxdong (approved):
Thanks for the back porting and the fix!

maluka-dmytro (approved):

PR #74, opened 2026-02-17
Pkvm v6.18: guest kernel patches

maluka-dmytro:
Port pKVM-IA guest kernel patches to v6.18.


maluka-dmytro:
Updated the patches to fully sync them with the patches we merged in https://android-review.googlesource.com/q/topic:%22pkvm-ia-6.18-guest%22. I think we should merge it.

Besides these patches, https://android-review.googlesource.com/q/topic:%22pkvm-ia-6.18-guest%22 also includes some strictly Android-specific patches, which I didn't include in this PR.

I was actually considering including at least the patch https://android-review.googlesource.com/c/kernel/common/+/3956313 (for disabling balloon for pVMs) together with the prior Android kernel patches it depends on, but it turns out I'd need to port quite a few of those Android patches, so I decided it isn't worth the trouble (and as I mentioned in the commit message of that patch, with non-Android 6.18 kernel balloon for pVMs is already effectively prevented anyway).

maluka-dmytro:
Also I'm gonna prepare a patch adding ia32_disable() in guest kernel (as discussed in https://github.com/intel-staging/pKVM-IA/issues/75), but maybe after we merge this PR.

cxdong (approved):
Yeah, makes sense for not including the balloon patch. And thanks for the updating!

PR #76, opened 2026-02-23
pKVM: x86: prevent hypervisor stack leaking

jaszczyk-grzegorz:
When pKVM handles host hypercalls it allocates an output structure on the stack, fills it accordingly during hypercall handling and finally use pkvm_hc_set_output() to map the data to the proper host vCPU registers so the host can read it back.

However, during hypercalls handling, some errors may occur that prevent the output data from being set. Because the pkvm_handle_host_hypercall() unconditionally copies the 'out' structure to the vCPU registers, uninitialized hyp stack (which potentially has sensitive data) can be leaked to the host.

To prevent described hypervisor stack leaking, zero initialize pkvm_hc_data 'out'.

Fixes: b0738f54947b ("pKVM: x86: Return hypercall outputs to the host")


maluka-dmytro (approved):

cxdong (approved):
Good cache!

PR #77, opened 2026-02-24
KVM: pVMX: Use fpu_user_cfg.max_size to compute fpsize

mmisono:
Since pKVM enforces the host's cpuids, calculate fpsize based on the value that the hardware supports instead of that of the vCPU reported.

Without this, pKVM fails to boot a VM on a machine with Intel AMX, as crosvm does not request that feature.

fixes: 58f48d1cb644b1de7b5c496191cc9d96b1f198b4 ("KVM: pVMX: Add new fpstate memory for xfd")

### DIFF
--- a/arch/x86/kvm/vmx/pkvm_host.c
+++ b/arch/x86/kvm/vmx/pkvm_host.c
@@ -1629,7 +1629,7 @@ static int pkvm_vcpu_realloc_fpstate(struct kvm_vcpu *vcpu)
 	void *fps;
 	int ret;
 
-	fpsize = PAGE_ALIGN(vcpu->arch.guest_fpu.fpstate->size +
+	fpsize = PAGE_ALIGN(fpu_user_cfg.max_size +
 			    ALIGN(offsetof(struct fpstate, regs), 64));
 	fps = alloc_pages_exact(fpsize, GFP_KERNEL_ACCOUNT);
 	if (!fps)
@@ -1663,7 +1663,7 @@ static void pkvm_vcpu_after_set_cpuid(struct kvm_vcpu *vcpu)
 	 * fpstate is only necessary for the pVM, and should be done before
 	 * adding the new cpuid entries to the pKVM hypervisor.
 	 */
-	if ((vcpu->arch.guest_fpu.xfeatures & XFEATURE_MASK_USER_DYNAMIC) &&
+	if ((fpu_user_cfg.max_features & XFEATURE_MASK_USER_DYNAMIC) &&
 	    pkvm_is_protected_vcpu(vcpu) &&
 	    pkvm_vcpu_realloc_fpstate(vcpu))
 		return;
--- a/arch/x86/kvm/cpuid.c
+++ b/arch/x86/kvm/cpuid.c
@@ -2138,6 +2138,38 @@ bool kvm_cpuid(struct kvm_vcpu *vcpu, u32 *eax, u32 *ebx,
 				*eax = vcpu->arch.hw_tsc_khz;
 			}
 #endif
+		} else if (pkvm_is_protected_vcpu(vcpu) && function == 0x15) {
+			/*
+			 * Since protected VMs cannot use kvmclock to find out the
+			 * TSC frequency, expose the native TSC frequency directly
+			 * via cpuid. For now this is not for security (pKVM does not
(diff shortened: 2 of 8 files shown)


arch/x86/kvm/vmx/pkvm_host.c
```diff
@@ -1629,7 +1629,7 @@ static int pkvm_vcpu_realloc_fpstate(struct kvm_vcpu *vcpu)
 	void *fps;
 	int ret;
 
-	fpsize = PAGE_ALIGN(vcpu->arch.guest_fpu.fpstate->size +
+	fpsize = PAGE_ALIGN(fpu_user_cfg.max_size +
```
cxdong, line 1632: Seems I didn't fully understand the root cause. I don't have a machine with Intel AMX feature, but suppose if the guest cpuid has enabled the XFD feature, the `vcpu->arch.guest_fpu.fpstate->size` will be set by the host KVM via `kvm_check_cpuid` -> `fpu_enable_guest_xfd_features` -> `__xfd_enable_feature` ->`fpstate_realloc`. Is this true from your side?

But as you mentioned in the commit message, the crosvm doesn't request the XFD feature for the guest, then  this function will not be called as `(vcpu->arch.guest_fpu.xfeatures & XFEATURE_MASK_USER_DYNAMIC) == false`.

mmisono: On my machine, [pkvm_vcpu_after_set_cpuid hypercallls](https://github.com/intel-staging/pKVM-IA/blob/de0cda16b6e2dc53a923ba3b0bc37d855538fd51/arch/x86/kvm/pkvm/pkvm.c#L1111) fails as __xfd_enable_feature() fails [here](https://github.com/intel-staging/pKVM-IA/blob/de0cda16b6e2dc53a923ba3b0bc37d855538fd51/arch/x86/kernel/fpu/xstate.c#L2060). AFAICT, in the hypercall handler, pkvm_enforce_cpuid() populate actual cpu's cpuid entries. Then `pkvm_vcpu_after_set_cpuid()` calls `kvm_set_cpuid() -> kvm_check_cpuid() -> fpu_enable_guest_xfd_features() -> __xfd_enable_feature()`, which triggers ENOMEM.


> then this function will not be called as (vcpu->arch.guest_fpu.xfeatures & XFEATURE_MASK_USER_DYNAMIC) == false.

yes, pkvm_vcpu_realloc_fpstate() is not called because of this. On my machine, I get:

```
[  138.684348] pkvm_host: [pkvm] vcpu->arch.guest_fpu.xfeatures & XFEATURE_MASK_USER_DYNAMIC = 0
[  138.684351] pkvm_host: [pkvm] vcpu->arch.guest_fpu.fpstate->size = 2560
[  138.684353] pkvm_host: [pkvm] fpu_user_cfg.max_features & XFEATURE_MASK_USER_DYNAMIC = 0x40000
[  138.684353] pkvm_host: [pkvm] fpu_user_cfg.max_size = 11008
```

cxdong: It looks like `pkvm_enforce_cpuid()` populate the actual cpu's cpuid entries, leaf 0xd and its subleaves, are not the same with the cpuid entries set by the crosvm?

If so, npVM should be fine?

mmisono: > It looks like pkvm_enforce_cpuid() populate the actual cpu's cpuid entries, leaf 0xd and its subleaves, are not the same with the cpuid entries set by the crosvm?

That is my understanding. Alternatively, enforce_cpuid() could respect crosvm's xfd configuration.

> If so, npVM should be fine?

My commit message is ambiguous but I have this issue for pVM. I just confirmed that npVM works fine without this change as you said.

cxdong: > That is my understanding. Alternatively, enforce_cpuid() could respect crosvm's xfd configuration.

This seems to be a better way.

>My commit message is ambiguous but I have this issue for pVM. I just confirmed that npVM works fine without this change as you said.

Thanks for confirming this.

mmisono: > This seems to be a better way.

as this patch fixes my issue, I don't plan work on this for the moment. Please feel free to discard/adopt this change in any way.

Issue #78, opened 2026-02-24
pkvm6.18 launch issue with enable_ipiv=0


mmisono:
I tried the latest PKVM 6.18 (commit: 86fdb9c05a31be5e) and found that a VM fails to boot if `kvm_intel.enable_apicv=0 kvm_intel.enable_ipiv=0` (if the CPU supports ipiv) because

- `enable_ipiv` is 1 in default: https://github.com/intel-staging/pKVM-IA/blob/de0cda16b6e2dc53a923ba3b0bc37d855538fd51/arch/x86/kvm/x86.c#L253
- pvkm-hyp does not use the module params currently, as written in TODO: https://github.com/intel-staging/pKVM-IA/blob/de0cda16b6e2dc53a923ba3b0bc37d855538fd51/arch/x86/kvm/vmx/vmx.c#L9990
- If `enable_ipiv=0` is given as a parameter, then the pkvm-host does not initialize `kvm_vmx->pid_table`, but pkvm-hyp still tries to use that: https://github.com/intel-staging/pKVM-IA/blob/de0cda16b6e2dc53a923ba3b0bc37d855538fd51/arch/x86/kvm/vmx/vmx.c#L5046-L5048

`kvm_intel.enable_apicv=1 kvm_intel.enable_ipiv=1` circumvent the issue.

cxdong:
Yeah, currently the pKVM hypervisor doesn't respect the host KVM's module parameter, for avoid exposing many local module parameters in the vmx.c. As the enable_apicv/enable_ipi are already exposed, we can just respect these two module parameters to fix this issue. Will submit a PR for this. Thanks!

PR #79, opened 2026-02-25
pKVM: VMX: Respect the host KVM's enable_apicv/enable_ipiv params

cxdong:
The host can boot with kvm_intel.enable_apicv=0, kvm_intel.enable_ipiv=0 or both. In this case, the enable_apicv/enable_ipiv from the pKVM hypervisor side still have the default true value, which makes the pKVM expects to host KVM to use apicv/ipiv and causes the failure to boot up a guest VM. Resolve this by making the pKVM's enable_apicv/enable_ipiv aligned with the host KVM's value before deprivilege.

Reported-by: Masanori Misono <m.misono760@gmail.com>
Closes: https://github.com/intel-staging/pKVM-IA/issues/78

### DIFF
--- a/arch/x86/include/asm/kvm_pkvm.h
+++ b/arch/x86/include/asm/kvm_pkvm.h
@@ -350,6 +350,8 @@ extern unsigned int pkvm_sym(tsc_khz);
 extern bool pkvm_sym(pvmfw_present);
 extern phys_addr_t pkvm_sym(pvmfw_base);
 extern phys_addr_t pkvm_sym(pvmfw_size);
+extern bool __read_mostly pkvm_sym(enable_apicv);
+extern bool __read_mostly pkvm_sym(enable_ipiv);
 
 u64 pkvm_total_reserve_pages(void);
 PKVM_DECLARE(void *, pkvm_early_alloc_page, (struct pkvm_memcache *mc));
--- a/arch/x86/include/asm/kvm_pkvm.h
+++ b/arch/x86/include/asm/kvm_pkvm.h
@@ -352,6 +352,7 @@ extern phys_addr_t pkvm_sym(pvmfw_base);
 extern phys_addr_t pkvm_sym(pvmfw_size);
 extern bool __read_mostly pkvm_sym(enable_apicv);
 extern bool __read_mostly pkvm_sym(enable_ipiv);
+extern bool __read_mostly pkvm_sym(enable_vpid);
 
 u64 pkvm_total_reserve_pages(void);
 PKVM_DECLARE(void *, pkvm_early_alloc_page, (struct pkvm_memcache *mc));
--- a/arch/x86/kvm/vmx/pkvm_init.c
+++ b/arch/x86/kvm/vmx/pkvm_init.c
@@ -118,6 +118,10 @@ static __init void pkvm_setup_syms(void)
 #endif
 	pkvm_sym(x86_pred_cmd) = x86_pred_cmd;
 	pkvm_sym(tsc_khz) = tsc_khz;
+
+	/* Respect below host KVM's module parameters */
+	pkvm_sym(enable_apicv) = enable_apicv;
+	pkvm_sym(enable_ipiv) = enable_ipiv;
 }
 
 static __init int pkvm_setup_host_vmcs_config(void)
--- a/arch/x86/kvm/vmx/pkvm_init.c
+++ b/arch/x86/kvm/vmx/pkvm_init.c
@@ -122,6 +122,7 @@ static __init void pkvm_setup_syms(void)
 	/* Respect below host KVM's module parameters */
 	pkvm_sym(enable_apicv) = enable_apicv;
 	pkvm_sym(enable_ipiv) = enable_ipiv;
+	pkvm_sym(enable_vpid) = enable_vpid;
 }
 
 static __init int pkvm_setup_host_vmcs_config(void)


arch/x86/kvm/vmx/pkvm_init.c
```diff
@@ -118,6 +118,10 @@ static __init void pkvm_setup_syms(void)
 #endif
 	pkvm_sym(x86_pred_cmd) = x86_pred_cmd;
 	pkvm_sym(tsc_khz) = tsc_khz;
+
+	/* Respect the host KVM's apicv/ipiv module parameters */
+	pkvm_sym(enable_apicv) = enable_apicv;
+	pkvm_sym(enable_ipiv) = enable_ipiv;
```
maluka-dmytro, line 124: Looks like it's easy enough to respect other parameters as well, not just these two, in the same way?

mmisono:
Functionality-wise, I confirmed that this fixes https://github.com/intel-staging/pKVM-IA/issues/78

arch/x86/kvm/vmx/pkvm_init.c
cxdong: It is easy, like we discussed before. The concern is if we want to respect all by this way, that may be a long list, and we also needs to expose the static ones. In the future if more module parameters are added, we also need to add it here. This looks like a bit heavy.

maluka-dmytro: Ah yeah, I forgot that many of those variables are static.

I've looked through `module_param*`s in KVM x86 code, and looks like indeed, among non-static ones, almost all are either not used by hypervisor code, or enforced by pKVM to specific values anyway. Looks like `enable_vpid` is an exception? so it might make sense to add just `enable_vpid` to this list here?

Also just to note: another non-static param `enable_device_posted_irqs` - currently not used by the hypervisor code (since posted interrupts are managed fully by the host), and currently pKVM doesn't support device assignment, but both things might change in the future?

cxdong: > so it might make sense to add just enable_vpid to this list here?

Yeah, adding one more already exposed module parameter should be easy. I will update.

> Also just to note: another non-static param enable_device_posted_irqs - currently not used by the hypervisor code (since posted interrupts are managed fully by the host), and currently pKVM doesn't support device assignment, but both things might change in the future?

Yeah, we can deal with this parameter then when we support the device assignment. At that time, we can see more clearly about if it is used by the hypervisor code or not.

cxdong: >>so it might make sense to add just enable_vpid to this list here?

>Yeah, adding one more already exposed module parameter should be easy. I will update.

Just a note: looks like the enable_vpid is not by the host's code (allocate_vpid/free_vpid/vmx_flush_tlb_all in vmx.c are not used), but is only used by the pKVM hypervisor's code. The major reason we want to respect the host's module parameter is that, those parameters are used by both side, and align them can fix some functional issues. But here for the enable_vpid, respecting the host's value just means there is a way to let the host configure this parameter for the pKVM hypervisor, which seems not that necessary?

maluka-dmytro: Yeah, it's just to let the user disable using VPID by the hypervisor, by setting the KVM param to false (since KVM allows that, and it's easy to be on par with KVM for that, so why not). E.g. just for testing/dev purposes, e.g. for experiments with performance.

And yeah, I don't imagine a really useful use case for that right now, so I'm fine with not adding it. OTOH, FWIW for example I've just looked up my old notes from my experiment with boot time performance of nested and double-nested VMs (without pKVM):

```
default setup:

L1: 10 seconds
L2: 20 seconds
L3: 2 minutes

with ept disabled in L0:

L1: 10 seconds
L2: 30 seconds
L3: >8 minutes

with shadow_vmcs disabled in L0:

L1: 10 seconds
L2: 25 seconds
L3: 4 minutes

with ept + shadow_vmcs disabled in L0:

L1: 10 seconds
L2: 40 seconds
L3: don't know, forever

with ept + shadow_vmcs + vpid disabled in L0:

L1: 10 seconds
L2: 55 seconds
L3: don't know, forever
```

i.e. showing that disabling vpid (in addition to disabling ept + shadow_vmcs) further slowed down nested VM bootup by 15 seconds, which is at least interesting. :)

cxdong: Makes sense. Will add enable_vpid.

arch/x86/include/asm/kvm_pkvm.h
```diff
@@ -350,6 +350,8 @@ extern unsigned int pkvm_sym(tsc_khz);
 extern bool pkvm_sym(pvmfw_present);
```
jaszczyk-grzegorz, line 350: nit to commit log: remove "Closes: #78" which is placed between Reported-by and Singed-off-by

jaszczyk-grzegorz (commented):
Othar than that LGTM

arch/x86/include/asm/kvm_pkvm.h
cxdong: Adding Closes: #78 is to resolve below checkpatch warning:
`WARNING: Reported-by: should be immediately followed by Closes: with a URL to the report`

So any reason for preferring removing this?

maluka-dmytro (approved):

arch/x86/include/asm/kvm_pkvm.h
jaszczyk-grzegorz: Ahh didn't know about such requirements and thought that this is leftover from PR description

jaszczyk-grzegorz (approved):

Issue #75, opened 2026-02-20
Protecting pkvm-ia from the interrupt injection issue


i-yyi:
Hi @maluka-dmytro,
I noticed that you added protection for L1 access to the VMCS in your latest version. In my recent research, I found that pkvm-ia does not seem to have added protection against interrupt injection yet.

In Security24, [Heckler](https://ahoi-attacks.github.io/heckler/) have been proposed. It's an attack method can break the cvm like get the ssh shell.

To my knowledge, pkvm now denied exception injection to pVM, but interrupts like 0x80 is allowed. Like __pkvm__interrupt_allowed, __pkvm__nmi_allowed. 

I'd like to ask if you've confirmed similar information, or if you're aware of the issue and have already implemented mitigation measures? If not, I'd like to help and submit a simple mitigation patch.

maluka-dmytro:
Thanks for looking into this. I think we haven't really looked into restricting interrupt injection yet. So if you submit a patch, I think it would be welcome. (Please note that the up-to-date branch is https://github.com/intel-staging/pKVM-IA/tree/pkvm-v6.18)

@cxdong maybe you or someone else from the Intel team have any more insight on this topic?

> Like __pkvm__interrupt_allowed, __pkvm__nmi_allowed.

FWIW I think __pkvm__interrupt_allowed and __pkvm__nmi_allowed have slightly different semantics: they are not for security checks, they are used for non-protected VMs as well, and they are, roughly speaking, a functional equivalent of vmx_interrupt_allowed() and vmx_nmi_allowed(). (BTW maybe these __pkvm__interrupt_allowed and __pkvm__nmi_allowed interfaces are temporary, since they are probably not quite optimal in terms of performance.)

I guess we'd just need to add some  checks to pkvm_inject_irq() for protected VMs.

cxdong:
[...]

PR #80, opened 2026-02-25
Pkvm v6.18 intr fix

cxdong:
Disable the software interrupt injection to the pVM, and validate the injected interrupt vector number is not in the range of 0 ~ 31.

### DIFF
--- a/arch/x86/kvm/pkvm/pkvm.c
+++ b/arch/x86/kvm/pkvm/pkvm.c
@@ -1803,7 +1803,8 @@ static int pkvm_vcpu_handle_host_hypercall(struct kvm_vcpu *hvcpu, enum pkvm_hc 
 void pkvm_handle_host_hypercall(struct kvm_vcpu *vcpu)
 {
 	enum pkvm_hc hc = pkvm_hc(vcpu);
-	union pkvm_hc_data in, out;
+	/* Zero 'out' to prevent leaking stack data on error */
+	union pkvm_hc_data in, out = {0};
 	int ret = 0;
 
 	pkvm_hc_get_input(vcpu, hc, &in);
--- a/arch/x86/kvm/pkvm/pkvm.c
+++ b/arch/x86/kvm/pkvm/pkvm.c
@@ -1930,19 +1930,6 @@ int pkvm_x86_vendor_init(struct kvm_x86_init_ops *ops)
 
 	memcpy(&kvm_x86_ops, ops->runtime_ops, sizeof(kvm_x86_ops));
 
-	if (!kvm_cpu_cap_has(X86_FEATURE_XSAVES))
-		kvm_caps.supported_xss = 0;
-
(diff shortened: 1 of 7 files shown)


arch/x86/kvm/pkvm/pkvm.c
```diff
@@ -1009,16 +1009,26 @@ static int pkvm_nmi_allowed(struct kvm_vcpu *vcpu, bool for_injection)
...
 }
 
-static void pkvm_inject_irq(struct kvm_vcpu *vcpu)
+static int pkvm_inject_irq(struct kvm_vcpu *vcpu)
 {
 	struct kvm_vcpu *shared_vcpu = to_pkvm_vcpu(vcpu)->shared_vcpu;
 
 	if (WARN_ON_ONCE(pkvm_interrupt_allowed(vcpu, true) <= 0))
-		return;
+		return -EINVAL;
```
maluka-dmytro: nit: maybe -EBUSY? Also, makes sense to do the same in pkvm_inject_nmi() as well?

arch/x86/kvm/pkvm/pkvm.c
```diff
@@ -1063,9 +1063,19 @@ static void pkvm_cancel_injection(struct kvm_vcpu *vcpu)
 		shared_vcpu->arch.nmi_injected = true;
 		vcpu->arch.nmi_injected = false;
 	} else if (vcpu->arch.interrupt.injected) {
-		kvm_queue_interrupt(shared_vcpu, vcpu->arch.interrupt.nr,
-				    vcpu->arch.interrupt.soft);
-		kvm_clear_interrupt_queue(vcpu);
+		/*
```
maluka-dmytro, line 1075: nit: commit message: s/Not to cancel/Do not cancel/ ?

arch/x86/kvm/pkvm/pkvm.c
cxdong: Sounds good. We probably need another patch for pkvm_inject_nmi()

arch/x86/kvm/pkvm/pkvm.c
cxdong: Sounds good to me.

maluka-dmytro (approved):

jaszczyk-grzegorz (approved):

cxdong:
@maluka-dmytro , you mentioned in https://github.com/intel-staging/pKVM-IA/issues/75#issuecomment-3977259618, we will have the protected apic page anyway, thus we don't need to have the workaround to disable apicv.

In this case, probably we also don't need to merge this PR right now, as the implementation of the apic page protection may change the PV interfaces modified by this PR, e.g., __pkvm__hwapic_isr_update/__pkvm__sync_pir_to_irr, again. So if we are not in a hurry of merging this PR, I would like to postpone it until I finish the apic page protection. WDYT?

maluka-dmytro:
SGTM.

PR #82, opened 2026-02-26
pKVM support for IOMMU

vineethrp:
This PR implements IOMMU management support in pKVM.

### DIFF
--- a/arch/x86/include/asm/kvm_host.h
+++ b/arch/x86/include/asm/kvm_host.h
@@ -2173,8 +2173,17 @@ void pkvm_create_vm_debugfs(struct kvm *kvm);
 int pkvm_vm_ioctl_enable_cap(struct kvm *kvm, struct kvm_enable_cap *cap);
 int kvm_topup_pkvm_memcache(struct pkvm_memcache *mc, unsigned long min_pages);
 void kvm_free_pkvm_memcache(struct pkvm_memcache *mc);
+
+DECLARE_STATIC_KEY_FALSE(pkvm_enabled_key);
+
+static inline bool pkvm_enabled(void)
+{
+	return static_branch_likely(&pkvm_enabled_key);
+}
 #else
 #define enable_pkvm		false
+static inline bool pkvm_enabled(void) { return false; }
+
 static inline void __init pkvm_reserve(void) {}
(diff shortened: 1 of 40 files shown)


cxdong:
Hi @vineethrp , just want to confirm this is the same patches we merged to Android17-6.18, right? If so, I will directly merge them :)

vineethrp:
> Hi @vineethrp , just want to confirm this is the same patches we merged to Android17-6.18, right? If so, I will directly merge them :)

Yeah this is merged to android17-6.18. I am fine with merging :-). @maluka-dmytro, what do you think?

maluka-dmytro:
Yeah, I believe that's our intention: just merge these patches as is, to align with android17-6.18.

BTW a small detail: this PR also includes Vineeth's patch https://github.com/intel-staging/pKVM-IA/pull/82/changes/325dd96e3a1138a239a9a968c388a84b51f36f8a from upstream, which was already automatically backported into android17-6.18 before we merged pvIOMMU.

In v6.12 this patch https://github.com/intel-staging/pKVM-IA/pull/82/changes/325dd96e3a1138a239a9a968c388a84b51f36f8a + next patch https://github.com/intel-staging/pKVM-IA/pull/82/changes/734588ffb58ba18f6118abdfcaaf6c4a7cc308ae were needed just to prevent a rarely reproducible issue (donation failure in case the pages allocated by the host kernel for donation are from the first 16MB of RAM). In v6.18, however, without these 2 patches, the system will not even boot (unless IOMMU passthrough mode is used), for the reason I described in https://github.com/intel-staging/pKVM-IA/pull/82/changes/734588ffb58ba18f6118abdfcaaf6c4a7cc308ae:

```
Dmytro: since v6.18 there is one more reason to ensure disabling
INTEL_IOMMU_FLOPPY_WA: pKVM for simplicity requires each DMA-mapped
range to be either fully normal memory or fully reserved memory, not a
mix of both, while those first 16MB are almost always a mix.
```

maluka-dmytro (approved):

maluka-dmytro:
> Yeah, I believe that's our intention: just merge these patches as is, to align with android17-6.18.

And I've just double-checked just in case that the diff of this PR matches the diff of the android17-6.18 patches. :)

Issue #87, opened 2026-03-20
pKVM protected VM setup fails after non-empty KVM_SET_CPUID2 on AMX-capable host


i-yyi:
Hi, I’m not sure whether you still remember an issue #42  I raised a long time ago about being unable to start a protected VM properly. At the time, the issue remained unresolved, and a colleague from Ant even helped reach out to Chuanxiao. The final conclusion then was that it had not been tested on Xeon, so I moved over to the Core platform instead.

A long time later — just a couple of days ago — I happened to try running it again on Xeon, considering that pkvm had changed significantly between 6.12 and 6.18. This time, I was surprised to find that crosvm no longer just hung indefinitely, but actually produced an error.

So I went back through the issue and found that it is related to fpstate introduced by AMX, and I also put together a simple patch.

Since I learned that you maintain the code on Gerrit rather than GitHub, I’m providing the patch file directly.

For detailed reproduction steps, please refer to the notes below.

On one x86 host, protected VM creation succeeds up to vCPU creation, but any non-empty `KVM_SET_CPUID2` causes subsequent pVM setup ioctls to fail. The same userspace flow works on another machine.

This reproduces both with CrosVM and with a minimal userspace probe that creates a pVM directly through KVM ioctls, which shows the issue is in the pKVM/KVM CPUID synchronization path rather than in CrosVM itself.

Environment

- Host kernel: `6.18.0+`
- VMM flow: QEMU boots Host0, then CrosVM starts Guest0 as a protected VM
- Affected host advertises dynamic user xfeatures (AMX-related XSAVE state)
- pKVM initialization completes successfully at boot

Observed behavior

- `KVM_CREATE_VM(KVM_X86_PKVM_PROTECTED_VM)` succeeds
- `KVM_CREATE_VCPU` succeeds
- `KVM_SET_CPUID2` with an empty CPUID list succeeds
- `KVM_SET_CPUID2` with any non-empty CPUID list succeeds initially, but the pVM is then effectively poisoned
- Subsequent ioctls such as `KVM_ENABLE_CAP(KVM_CAP_X86_PROTECTED_VM, INFO)` and `KVM_SET_USER_MEMORY_REGION` fail with `-EIO`
- CrosVM cannot finish Guest0 startup

Minimal reproduction

A minimal probe showed the following behavior:

- `create-vm`: PASS
- `create-vcpu`: PASS
- `memslot-only`: PASS
- `empty-cpuid`: PASS
[...]

PR #88, opened 2026-03-20
KVM: x86: pKVM: align AMX CPUID with host-provided model

i-yyi:
This is a vibe-coding version for #87 . If you think it generally matches what you had in mind, I can do the first round of review myself. Once I’ve confirmed it, we can move on to the next round of discussion.

----
For protected VMs, pKVM enforces a host-like CPUID and may append missing CPUID leaves from the default set. On AMX-capable hosts, that can expose AMX-related CPUID state even when the host userspace VMM didn't provide AMX in the guest CPUID model.

That mismatch is problematic in two ways. First, it changes the guest CPU model behind the VMM's back. Second, when AMX state is added during pKVM enforcement, the hypervisor may require a larger fpstate buffer than what the host side prepared for the protected vCPU.

Align pKVM's AMX handling with the CPUID model provided by the host VMM:

- if the host-provided CPUID contains AMX, keep the AMX-related CPUID bits/leaves and reallocate fpstate before synchronizing CPUID into the hypervisor;
- if the host-provided CPUID does not contain AMX, clear the AMX-related CPUID state from the enforced result and refresh xstate sizing.

This keeps AMX exposure under the VMM's control while preserving the required fpstate sizing when AMX is intentionally exposed.

### DIFF
--- a/arch/x86/kvm/cpuid.c
+++ b/arch/x86/kvm/cpuid.c
@@ -2303,8 +2303,130 @@ static void pkvm_fixup_cpuid_entry(struct kvm_cpuid_entry2 *entry)
 #define CPUID_1_ECX_TSC_DLTIMER		(1 << 24)
 #define CPUID_1_ECX_HYP			(1 << 31)
 #define CPUID_1_EDX_HTT			(1 << 28)
+static bool cpuid_entry_is_empty(struct kvm_cpuid_entry2 *e2);
+
+static bool pkvm_cpuid_entry_has_amx(struct kvm_cpuid_entry2 *entry)
+{
+	u64 xfeatures;
+
+	switch (entry->function) {
+	case 7:
+		if (!entry->index)
+			return cpuid_entry_has(entry, X86_FEATURE_AMX_TILE) ||
+			       cpuid_entry_has(entry, X86_FEATURE_AMX_INT8) ||
+			       cpuid_entry_has(entry, X86_FEATURE_AMX_BF16);
+		if (entry->index == 1)
+			return cpuid_entry_has(entry, X86_FEATURE_AMX_FP16) ||
+			       cpuid_entry_has(entry, X86_FEATURE_AMX_COMPLEX);
+		return false;
+	case 0xd:
+		if (entry->index == 0) {
+			xfeatures = entry->eax | ((u64)entry->edx << 32);
+			return xfeatures & XFEATURE_MASK_XTILE;
+		}
+
+		return entry->index == XFEATURE_XTILE_CFG ||
+		       entry->index == XFEATURE_XTILE_DATA;
+	case 0x1d:
+	case 0x1e:
+		return entry->function || entry->eax || entry->ebx ||
+		       entry->ecx || entry->edx;
+	default:
+		return false;
+	}
+}
+
+static bool pkvm_cpuid_has_amx(struct kvm_cpuid_entry2 *e2, int nent)
+{
+	int i;
+
+	for (i = 0; i < nent; i++) {
+		if (cpuid_entry_is_empty(&e2[i]))
+			continue;
+		if (pkvm_cpuid_entry_has_amx(&e2[i]))
+			return true;
+	}
+
+	return false;
+}
+
+static void pkvm_cpuid_align_amx(struct kvm_cpuid_entry2 *entry,
+				 struct kvm_cpuid_entry2 *host,
+				 bool has_amx)
+{
+	u64 xfeatures;
+
+	switch (entry->function) {
+	case 7:
+		if (!entry->index) {
+			bool keep_tile = host && cpuid_entry_has(host, X86_FEATURE_AMX_TILE);
+			bool keep_int8 = host && cpuid_entry_has(host, X86_FEATURE_AMX_INT8);
+			bool keep_bf16 = host && cpuid_entry_has(host, X86_FEATURE_AMX_BF16);
+
+			cpuid_entry_change(entry, X86_FEATURE_AMX_TILE, keep_tile);
+			cpuid_entry_change(entry, X86_FEATURE_AMX_INT8, keep_int8);
+			cpuid_entry_change(entry, X86_FEATURE_AMX_BF16, keep_bf16);
+		} else if (entry->index == 1) {
+			bool keep_fp16 = host && cpuid_entry_has(host, X86_FEATURE_AMX_FP16);
+			bool keep_complex = host && cpuid_entry_has(host, X86_FEATURE_AMX_COMPLEX);
+
+			cpuid_entry_change(entry, X86_FEATURE_AMX_FP16, keep_fp16);
+			cpuid_entry_change(entry, X86_FEATURE_AMX_COMPLEX, keep_complex);
+		}
+		break;
+	case 0xd:
+		if (!has_amx) {
+			if (entry->index == 0) {
+				xfeatures = entry->eax | ((u64)entry->edx << 32);
+				xfeatures &= ~XFEATURE_MASK_XTILE;
+				entry->eax = xfeatures;
+				entry->edx = xfeatures >> 32;
+				entry->ebx = xstate_required_size(xfeatures, false);
+				entry->ecx = entry->ebx;
+			} else if (entry->index == XFEATURE_XTILE_CFG ||
+				   entry->index == XFEATURE_XTILE_DATA) {
+				memset(entry, 0, sizeof(*entry));
+			}
+		}
+		break;
+	case 0x1d:
+	case 0x1e:
+		if (!has_amx)
+			memset(entry, 0, sizeof(*entry));
+		break;
+	default:
+		break;
+	}
+}
+
+static void pkvm_refresh_xstate_cpuid(struct kvm_cpuid_entry2 *e2, int nent)
+{
+	struct kvm_cpuid_entry2 *d0, *d1;
+	u64 xcr0, xss;
+
+	d0 = kvm_find_cpuid_entry2(e2, nent, 0xd, 0);
+	if (!d0)
+		return;
+
+	xcr0 = d0->eax | ((u64)d0->edx << 32);
+	d0->ebx = xstate_required_size(xcr0, false);
+	d0->ecx = d0->ebx;
+
+	d1 = kvm_find_cpuid_entry2(e2, nent, 0xd, 1);
+	if (!d1)
+		return;
+
+	xss = d1->ecx | ((u64)d1->edx << 32);
+	if (cpuid_entry_has(d1, X86_FEATURE_XSAVES) ||
+	    cpuid_entry_has(d1, X86_FEATURE_XSAVEC))
+		d1->ebx = xstate_required_size(xcr0 | xss, true);
+	else
+		d1->ebx = 0;
+}
+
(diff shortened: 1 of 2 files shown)


cxdong:
> This is a vibe-coding version for https://github.com/intel-staging/pKVM-IA/issues/87 . If you think it generally matches what you had in mind, I can do the first round of review myself. Once I’ve confirmed it, we can move on to the next round of discussion.

Basically it is. I have posted my understanding about the root cause in https://github.com/intel-staging/pKVM-IA/issues/87#issuecomment-4095352098. With this understanding, I think we may not need to modify the pkvm_host.c.

Specifically, we may only need to align the `XFEATURE_MASK_USER_DYNAMIC` bits in the CPUID 0xd  ECX 0 (refer to [kvm_check_cpuid](https://github.com/intel-staging/pKVM-IA/blob/pkvm-v6.18/arch/x86/kvm/cpuid.c#L175)). If the CPUID from host VMM doesn't have `XFEATURE_MASK_USER_DYNAMIC` bits, then make sure the `pkvm_enforce_cpuid` won't add these bits. We may also needs to update the size information stored in the CPUID 0xd main and sub leaves. The size calculation can refer to [__do_cpuid_func](https://github.com/intel-staging/pKVM-IA/blob/pkvm-v6.18/arch/x86/kvm/cpuid.c#L1568).

What do you think?

i-yyi:
I understand, I agree with your point of view, and I will complete this PR during my free time next week.

maluka-dmytro:
> For protected VMs, pKVM enforces a host-like CPUID and may append missing CPUID leaves from the default set. On AMX-capable hosts, that can expose AMX-related CPUID state even when the host userspace VMM didn't provide AMX in the guest CPUID model.

Am I correct that this problem exists only in the case when the VMM doesn't use the KVM_GET_SUPPORTED_CPUID ioctl for creating the guest CPUID (or uses it but for some reason removes AMX stuff from it)?

But in https://github.com/intel-staging/pKVM-IA/issues/87 you mentioned you observed this with crosvm, whereas crosvm does use KVM_GET_SUPPORTED_CPUID, and it doesn't look like crosvm removes AMX stuff. What am I missing?

> That mismatch is problematic in two ways. First, it changes the guest CPU model behind the VMM's back.

Yeah, I agree this is not quite nice. Basically it's a hack (BTW together with the optimistic assumption that the aligned buffer will be large enough for extra entries).

As I see it, a proper solution would be: pKVM _requires_ the VMM to prepare a CPUID matching pKVM's requirements, and _validates_ this CPUID and returns an error if it doesn't conform, instead of silently modifying it. IIUC that is more or less how it works in TDX. (The requirements themselves might be the same as what pKVM already enforces, i.e. roughly speaking: based on KVM_GET_SUPPORTED_CPUID, with only a few leaves allowed to be tweaked by the VMM.) 

I think basically the reason why we didn't already implement it this way was just to spare the effort, and spare the need to modify crosvm for now...

i-yyi:
> Am I correct that this problem exists only in the case when the VMM doesn't use the KVM_GET_SUPPORTED_CPUID ioctl for creating the guest CPUID (or uses it but for some reason removes AMX stuff from it)?
But in https://github.com/intel-staging/pKVM-IA/issues/87 you mentioned you observed this with crosvm, whereas crosvm does use KVM_GET_SUPPORTED_CPUID, and it doesn't look like crosvm removes AMX stuff. What am I missing?

I don't think so. My understanding is that for protected VMs, even if the VMM (such as crosvm) calls KVM_GET_SUPPORTED_CPUID and provides appropriate settings, PKVM will not obey the CPUID provided by crosvm, but will instead overwrite it according to the host CPU.
If I am wrong, please point out my mistake.

> As I see it, a proper solution would be: pKVM requires the VMM to prepare a CPUID matching pKVM's requirements, and validates this CPUID and returns an error if it doesn't conform, instead of silently modifying it. IIUC that is more or less how it works in TDX. (The requirements themselves might be the same as what pKVM already enforces, i.e. roughly speaking: based on KVM_GET_SUPPORTED_CPUID, with only a few leaves allowed to be tweaked by the VMM.)

Yes, PKVM is better suited as a verification gate than for direct intervention. I agree with you.

cxdong:
> Am I correct that this problem exists only in the case when the VMM doesn't use the KVM_GET_SUPPORTED_CPUID ioctl for creating the guest CPUID (or uses it but for some reason removes AMX stuff from it)?

I guess the CPUID bits are not removed by the crosvm. When the host KVM return the supported CPUID for the  KVM_GET_SUPPORTED_CPUID ioctl, the host KVM has removed the XTILE_DATA bit from the permitted_xcr0 via the `kvm_get_filtered_xcr0()` as the crosvm process is not requesting to use it. But from the pKVM side, as there is no way for the pKVM to know about the crosvm process's permitted XCR0, thus XTILE_DATA bit is retained, which made the difference when calculating the FPU memory size.

AFAIK, the linux kernel made such mechanism to let the process explicitly request to use AMX feature rather than allow it by default is to reduce the memory consumption from the FPU state for each process, as not every process will need AMX.

> As I see it, a proper solution would be: pKVM requires the VMM to prepare a CPUID matching pKVM's requirements, and validates this CPUID and returns an error if it doesn't conform, instead of silently modifying it. IIUC that is more or less how it works in TDX. (The requirements themselves might be the same as what pKVM already enforces, i.e. roughly speaking: based on KVM_GET_SUPPORTED_CPUID, with only a few leaves allowed to be tweaked by the VMM.)
>
>I think basically the reason why we didn't already implement it this way was just to spare the effort, and spare the need to modify crosvm for now...

Yes, true. That is the reason. With the current CPUID enforcment mechanism, I think the XTILE_DATA could be an allowed bit for the host to tweak, so that the pKVM can horner this bit from the host (e.g., in this case, remove XTILE_DATA from the pKVM's side and update the size information in the CPUID as well).

cxdong:
> But from the pKVM side, as there is no way for the pKVM to know about the crosvm process's permitted XCR0, thus XTILE_DATA bit is retained, which made the difference when calculating the FPU memory size.

This reminds me that, a possible alternative is to let the pKVM respect the host KVM's guest FPU permission, which is used to calculate the permitted XCR0 in `xstate_get_group_perm()`, so that when the pKVM calculate the permitted XCR0, it can get the same result with the host. Currently the guest FPU permission from the pKVM side is fixed, thus there is no per-VM one, which may be needed in this alternative.

maluka-dmytro:
> When the host KVM return the supported CPUID for the KVM_GET_SUPPORTED_CPUID ioctl, the host KVM has removed the XTILE_DATA bit from the permitted_xcr0 via the `kvm_get_filtered_xcr0()` as the crosvm process is not requesting to use it. But from the pKVM side, as there is no way for the pKVM to know about the crosvm process's permitted XCR0, thus XTILE_DATA bit is retained, which made the difference when calculating the FPU memory size.
> 
> AFAIK, the linux kernel made such mechanism to let the process explicitly request to use AMX feature rather than allow it by default is to reduce the memory consumption from the FPU state for each process, as not every process will need AMX.

I see. So in a nutshell, the problem is that this is per userspace process, and pKVM currently has no knowledge about such host's per-process states.

Also, this problem seems quite independent of how pKVM enforces the CPUID - by silently modifying it (like it does now) or just by validating it? (i.e. in the latter case we'd still need to address this problem somehow?)

I'm not really familiar with all this FPU stuff, but it feels it would be better to address this problem in some generic way, not by adding ad-hoc logic for this specific AMX issue...

maluka-dmytro:
BTW just for the record: not sure if exactly related to this issue, but I recall that when originally reviewing Kevin's patch in https://android-review.googlesource.com/c/kernel/common/+/3813637, I had the following observation:

> On all vCPUs, pKVM enforces different values of some FPU params (according to SDM, those params are "Maximum size (bytes, from the beginning of the XSAVE/XRSTOR save area) required by
enabled features in XCR0" and "supported bits of the lower 32 bits of the IA32_XSS MSR"):
> ```
> [  230.485715] pkvm: orig cpuid: func=d idx=0 flags=1 eax=207 ebx=240 ecx=a88 edx=0
> [  230.493979] pkvm:  new cpuid: func=d idx=0 flags=1 eax=207 ebx=a88 ecx=a88 edx=0
> [  230.502230] pkvm: orig cpuid: func=d idx=1 flags=1 eax=f ebx=240 ecx=0 edx=0
> [  230.510102] pkvm:  new cpuid: func=d idx=1 flags=1 eax=f ebx=348 ecx=0 edx=0
> ```
> I haven't analyzed if it is ok that pKVM changes these values. So, is it ok?

And Kevin's reply was:

> FPU is managed by pKVM hence the state size must be set by pKVM (e.g. it doesn't sound secure if the host reports a smaller area size than supported FPU features to guest, leading to xsave/xrstor not covering the full state in guest). In this specific case the difference comes from:
> 
> a) crosvm gets size a88 from kvm-high, based on supported FPU features
> b) crosvm passes a88 to kvm-high
> c) kvm-high updates it to 240 based on initial xcr0 value. See kvm_set_cpuid() calls __kvm_update_cpuid_runtime()
> d) kvm-high issues hypercall with 240 to pkvm, at the end of kvm_set_cpuid()
> e) pkvm calls pkvm_enforce_cpuid(), which updates 240 to a88 based on supported FPU features
> f) pkvm calls kvm_set_cpuid(), same as step c). Then 240 again
> 
> at run-time, guest kernel updates xcr0, emulated by __kvm_set_xcr() which calls kvm_update_cpuid_runtime() to change 240 to a88 based on the new xcr0 value.
> 
> So in the end the guest observes the desired value.

cxdong:
> I see. So in a nutshell, the problem is that this is per userspace process, and pKVM currently has no knowledge about such host's per-process states.

Yes it is. So I raised the alternative in https://github.com/intel-staging/pKVM-IA/pull/88#issuecomment-4124047904, which is to sync the guest vCPU process's fpu permission to the pKVM. With this, the pKVM will also remove the `XFEATURE_XTILE_DATA` bit from its side if this is not requested by the host. This alternative doesn't need to modify the current CPUID enforcement logic.

> Also, this problem seems quite independent of how pKVM enforces the CPUID - by silently modifying it (like it does now) or just by validating it? (i.e. in the latter case we'd still need to address this problem somehow?)
>
>I'm not really familiar with all this FPU stuff, but it feels it would be better to address this problem in some generic way, not by adding ad-hoc logic for this specific AMX issue...

If the pKVM only validates rather than silently modifies the CPUID bits, and returns error code to the host, then the host should guarantee only exposing the expected CPUID bits to the pKVM to avoid the validation failure. This is the solution you mentioned in https://github.com/intel-staging/pKVM-IA/pull/88#issuecomment-4122806339, right? Without the above alternative, the crosvm should request the `XFEATURE_XTILE_DATA` fpu permission for the guest vCPU process, then the host KVM can expose the CPUID which contains the `XFEATURE_XTILE_DATA` bit required by the pKVM, which seems not necessary if the crosvm want to create a guest which doesn't need the AMX feature. With the above alternative, the crosvm doesn't need to do this as the `XFEATURE_XTILE_DATA` bit will be tweaked by the pKVM according to the guest vCPU process's fpu permission.

So seems this alternative is needed for either for the solution in https://github.com/intel-staging/pKVM-IA/pull/88#issuecomment-4122806339 or for the current CPUID enforcement.

PR #90, opened 2026-03-23
SQUASHME: pKVM: x86: Fix vmexit trace dump lockup

cxdong:
The vmexit trace dump path can return -ENOSPC from the hypervisor while holding vm->lock in __copy_guest_vm_trace(), which leaves the per-VM lock permanently held and can strand a CPU in hyp with interrupts disabled.

Fix it by always dropping vm->lock on the error path.

Fixes: 1ff5d9013996 ("pKVM: x86: trace: Extend PV interface to dump the guest trace")

### DIFF
--- a/arch/x86/kvm/pkvm/trace.c
+++ b/arch/x86/kvm/pkvm/trace.c
@@ -100,8 +100,10 @@ static int __copy_guest_vm_trace(struct pkvm_vm *vm, void *param)
 		if (!vm->vcpus[i])
 			continue;
 
-		if (arg->size < sizeof(struct perf_data))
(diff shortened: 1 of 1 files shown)


maluka-dmytro (approved):

maluka-dmytro:
Thanks for picking this up.

There are also a few other small fixes to port from android17-6.18, so I've created PR https://github.com/intel-staging/pKVM-IA/pull/91 with them.

cxdong:
Thanks! The PR https://github.com/intel-staging/pKVM-IA/pull/91 is merged.

PR #91, opened 2026-03-23
Port some IOMMU fixes/improvements from android17-6.18

maluka-dmytro:
Port a few recent fixes & improvements:

- Fix missing error check for cache_tag_assign_domain() in passthrough mode, which was causing a potential security issue due to lack of TLB flush.

- Avoid disabling IOMMU if `iommu=off` or `intel_iommu=off` is given in the kernel command line (if pKVM is enabled).

- Avoid disabling GPU's IOMMU if `intel_iommu=igfx_off` is given in the kernel command line (if pKVM is enabled). As a bonus, repurpose `intel_iommu=igfx_off` if pKVM is enabled: instead of completely disabling the GPU's IOMMU, set IOMMU passthrough mode for the GPU only (which is useful e.g. for experiments with GPU performance).

### DIFF
--- a/drivers/iommu/intel/iommu.c
+++ b/drivers/iommu/intel/iommu.c
@@ -47,6 +47,8 @@
 #define CONTEXT_SIZE		VTD_PAGE_SIZE
 
 #define IS_GFX_DEVICE(pdev) pci_is_display(pdev)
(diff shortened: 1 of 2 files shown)


vineethrp:
LGTM

cxdong:
Thanks for porting this to github.

Issue #92, opened 2026-03-24
pkvm_domain_map fail in 6.18 pviommu


mmisono:
I tried the latest 6.18 (e091271f109042d2c224b747ac4c48950c457016) and had some issues with pvIOMMU on my machine.

First, the machine has 17 DRHDs, so `PKVM_MAX_IOMMU_NUM` needs to be increased. 

Secondly, during the boot,
```
[   14.551644] DMAR: pkvm: pkvm_domain_map: domain map[iov_pfn: 0, pfn: 0, nr_pages: 4096] failed (err=-22)
[   14.561208] iommu dmar17: IOMMU driver was not able to establish FW requested direct mapping.
```
Apparently, it fails to create the IOMMU mapping that ACPI RMRR requested (because the region is a mix of normal and reserved regions, as written in `pkvm_host_use_dma()`). The pKVM initialization continues and the machine boots anyway, and it seems ok not to have this mapping in my machine. 

The real problem happens when I run some experiments with virtio-blk. I couldn't make this fully reproducible, but it seems this happens when a protected VM has a virtio-blk with O_DIRECT (`-block disk.img,direct=true`). With such a configuration, during a pVM boot (I added some dbg messages),

```
[  138.624083] [pkvm] page->host_state=3, expected states mask=0x2
[  138.624096] [pkvm] pkvm_host_use_dma: invalid memory range [0x2a1b9c000, 0x2a1b9d000), state check failed with -1
[  138.624101] DMAR: pkvm: pkvm_domain_map: domain map[iov_pfn: fdf00, pfn: 2a1b9c, nr_pages: 1] failed (err=-1)
[  138.634193] DMAR: DRHD: handling fault status reg 3
[  138.635163] [pkvm] page->host_state=3, expected states mask=0x2
[  138.639176] [pkvm] pkvm_host_use_dma: invalid memory range [0x2a1b9c000, 0x2a1b9d000), state check failed with -1
[  138.639180] DMAR: [DMA Write NO_PASID] Request device [b1:00.0] fault addr 0xfdf00000 [fault reason 0x71] SM: Present bit in first-level paging entry is clear
[  138.639193] DMAR: pkvm: pkvm_domain_map: domain map[iov_pfn: fdf00, pfn: 2a1b9c, nr_pages: 1] failed (err=-1)
[  138.663508] DMAR: DRHD: handling fault status reg 3
[...]

PR #93, opened 2026-03-25
SQUASHME: pKVM: x86: Fix dependency on !BLK_DEV_FD

maluka-dmytro:
!BLK_DEV_FD doesn't cover the CONFIG_BLK_DEV_FD=m case, it only covers the CONFIG_BLK_DEV_FD=y case.
See Documentation/kbuild/kconfig-language.rst:

>            '!' <expr>                           (6)
>
> ...
>
> (6) Returns the result of (2-/expr/).
>
> ...
>
> An expression can have a value of 'n', 'm' or 'y' (or 0, 1, 2
> respectively for calculations).

So !BLK_DEV_FD doesn't prevent enabling pKVM together with the floppy driver if the floppy driver is enabled as a module.

So change it to BLK_DEV_FD=n to prevent both =y and =m cases.

Fixes: 38ed4e8f61ec ("pKVM: x86: set PKVM_INTEL depend on !BLK_DEV_FD")
Reported-by: Masanori Misono <m.misono760@gmail.com>
Link: https://github.com/intel-staging/pKVM-IA/issues/92

### DIFF
--- a/arch/x86/kvm/Kconfig
+++ b/arch/x86/kvm/Kconfig
@@ -138,7 +138,7 @@ config PKVM_INTEL
 	bool "pKVM for Intel processors support"
 	depends on KVM_INTEL=y
 	depends on INTEL_IOMMU
-	depends on !BLK_DEV_FD
+	depends on BLK_DEV_FD=n
 	select PKVM_X86
 	help
 	  Provides support for pKVM on Intel processors.


maluka-dmytro:
@vineethrp FYI

vineethrp:
> @vineethrp FYI

LGTM

cxdong (approved):

PR #98, opened 2026-04-15
x86/pkvm: Add dependency with CONFIG_PARAVIRT

cxdong:
As the pKVM guest kernel will hook its own mmio operations to the pv_ops, add dependency with CONFIG_PARAVIRT to make sure the pv_ops is enabled.

Fixes: 056f607f79d1 ("x86/pkvm: Use hypercalls to access MMIO")
Reported-by: kernel test robot <lkp@intel.com>


cxdong:
Back ported from https://android.googlesource.com/kernel/common/+/refs/heads/android17-6.18

Issue #104, opened 2026-05-16
Potential isolation violation due to hugepage unmap.


i-yyi:
Hi there,
I've recently found another new issue in pkvm's iommu driver. When host maliciously unmap a basepage in a hugepage , instead of rejecting or splitting the hugepage, the walker treats the superpage data HPA as a lower level page table, can clear host-controlled fake DMA PTEs, and can unpin a still-present DMA-mapped host page that is later donated to a protected VM. Which make the host can fully control the guest memory.

You may trigger this bug as follow steps :
1. prepare a 2M hugepage, called superpage
2. select a 4K page in the superpage, called targetpage
3. write a fake pte that point to targetpage at superpage+8
4. iommu_map that superpage
5. use hypercall (rather than interface in the host kernel) to unmap the targetpage
6. this will result-> targetpage refcount decreased, and may donate to a pvm in the future, however ,the map of the superpage still present in the IOPT.

I'm not sure if my description is clear. If any additional information is needed or the vulnerability is invalid, please let me know at any time. Thank you.

maluka-dmytro:
I assume you are referring to the code in `dma_pte_clear_level()`.

At first glance looks like we indeed have an issue there, since before recursively descending to the next level, it doesn't check if the PTE at the current level doesn't point to a hugepage (i.e. doesn't check `!dma_pte_superpage(pte)`).

@vineethrp @aashishsharma7 what do you think?

BTW in newer upstream kernels (since commit d373449d8e97 ("iommu/vt-d: Use the generic iommu page table")) that code has been thrown away and replaced with the generic IOPT code in `drivers/iommu/generic_pt/iommu_pt.h`, and at first glance it looks like that generic code does properly check for the hugepage bit before recursively descending to the next level: in `__unmap_range()`:

```
		if (pts.type != PT_ENTRY_OA) {
			bool fully_covered;

			if (pts.type != PT_ENTRY_TABLE) {
				ret = -EINVAL;
				break;
			}

...
			ret = pt_descend(&pts, arg, __unmap_range);
			if (ret)
				break;
...
```

Also BTW there is the following comment in that generic IOPT code:
```
	/*
	 * A starting index is in the middle of a contiguous entry
	 *
	 * The IOMMU API does not require drivers to support unmapping parts of
	 * large pages. Long ago VFIO would try to split maps but the current
	 * version never does.
	 *
	 * Instead when unmap reaches a partial unmap of the start of a large
[...]

PR #105, opened 2026-05-22
pKVM: iommu/vt-d: Reject partial unmap of superpage

vineethrp:
The pKVM IOMMU unmap hypercall receives its PFN range from the host, so it cannot rely on the normal host IOMMU path to have expanded partial unmaps of large mappings. If a partial range crosses a superpage leaf, the shared unmap walker can otherwise descend through the leaf and interpret DMA data page as a lower-level page table.

Add defensive checks before recursive descent so host/pKVM does not treat a superpage leaf as a page-table pointer.

Reported-by: Yi Ran <yyi@stu.pku.edu.cn>
Closes: https://github.com/intel-staging/pKVM-IA/issues/104
Fixes: 119a4e6a8273 ("iommu/vt-d: pKVM: Hypercalls for domain pagetable management")

### DIFF
--- a/drivers/iommu/intel/iommu.c
+++ b/drivers/iommu/intel/iommu.c
@@ -1008,6 +1008,9 @@ static bool dma_pte_clear_level(struct dmar_domain *domain, int level,
 				first_pte = pte;
 			last_pte = pte;
 		} else if (level > 1) {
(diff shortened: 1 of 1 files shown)


maluka-dmytro:
This patch has been already merged as a part of https://github.com/intel-staging/pKVM-IA/pull/106 so we can close this PR?

PR #109, opened 2026-06-03
pKVM improvements & aosp aligment

jaszczyk-grzegorz:
This PR is to align with https://android-review.googlesource.com/c/kernel/common/+/4092092 where those patches got merged.


maluka-dmytro:
Remove `Change-Id:` tags?

cxdong:
Thanks for backporting! I agree that it is better to remove `Change-Id:` to align with the other merged patches.

jaszczyk-grzegorz:
Sure, sorry forgot about it - removed.

PR #110, opened 2026-06-04
Harden host access to IOMMU MMIO registers

vineethrp:
pKVM passes the IOMMU through to the host but traps its MMIO register accesses. This series hardens those accesses with a deny-by-default model: only the registers the current host driver actually uses are allowed through, and every allowed write is validated against the VT-d register semantics (reserved/read-only fields, value constraints) so a malicious or buggy host cannot violate pKVM's security guarantees or drive the hardware into undefined behavior.


cxdong:
Thanks for back porting this!

PR #115, opened 2026-08-01
iommu/vt-d: Backport pKVM security fixes

vineethrp:
## Summary

Backport 23 reviewed security and correctness fixes for the protected Intel
VT-d implementation, including PASID/context teardown serialization, IOMMU
domain lifetime, translation controls, cache-tag handling, and validation of
host-supplied domain state.

## Source

- Android common Gerrit series:
  https://android-review.googlesource.com/c/kernel/common/+/4172676

## VT-d prerequisite backports

The following two commits were cherry-picked from the upstream
`linux-6.18.y` stable tree before the 23-patch pKVM IOMMU series. They align
the PASID/context teardown logic with the source series and avoid conflicts
while applying its IOMMU fixes. The same commit objects also appear in the
Android common tree, so their object IDs are identical in both repositories:

- `821807c167b7` — iommu/vt-d: Clear Present bit before tearing down PASID entry
  https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/commit/?h=linux-6.18.y&id=821807c167b7
- `d2138abc8f0a` — iommu/vt-d: Clear Present bit before tearing down context entry
  https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/commit/?h=linux-6.18.y&id=d2138abc8f0a

## Validation

- Replayed all 25 commits independently onto
  `intel-staging/pKVM-IA:pkvm-v6.18`.
- Built successfully.
- Verified every `Fixes:` target is an earlier commit with the quoted subject.
- Verified clean pairwise merges with the other two backport series.

### DIFF
--- a/arch/x86/include/asm/kvm_pkvm.h
+++ b/arch/x86/include/asm/kvm_pkvm.h
@@ -93,12 +93,12 @@ struct pasid_setup_fl_data {
 	u64 fsptptr_gpa;
 	u64 donation_page_gpa;
 	u32 pasid;
-	u32 flags;
 	u16 did;
 	u8 bus;
 	u8 devfn;
 	u8 ats_qdep;
 	u8 ats_supported: 1;
+	u8 force_snoop: 1;
 };
 
 struct pasid_setup_sl_data {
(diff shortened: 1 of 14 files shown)


cxdong:
Thanks for these backport PRs!

PR #113, opened 2026-08-01
pKVM: Backport VMX host-state security fixes

vineethrp:
## Summary

Backport seven reviewed pKVM/VMX security fixes for host register state,
exception fixup, PKRU handling, XCR0/XSS protection, and user-return MSR
handling.

## Source

- Android common Gerrit series:
  https://android-review.googlesource.com/c/kernel/common/+/4175821

## Validation

- Replayed independently onto `intel-staging/pKVM-IA:pkvm-v6.18`.
- Built successfully.
- Verified every `Fixes:` target is an earlier commit with the quoted subject.
- Verified clean pairwise merges with the other two backport series.

PR #116, opened 2026-08-02
pKVM: VMX: Handle unregistered user-return MSRs

vineethrp:
## Summary

Adding a patch for the changes I missed in PR #113 

This correction was included in revision 13 of the Android kernel change, but
was missed when the earlier revision was ported to the Intel tree. It follows
up merged commit `3bd17dc24aba` (`pKVM: VMX: Intercept host writes to
user-return MSRs`).

Android source:
https://android-review.googlesource.com/c/kernel/common/+/4175821/13

## Validation

- verified that the code matches the revision-13 delta exactly
- `git diff --check`
- `scripts/checkpatch.pl --strict -g HEAD`
- Build succesfully


maluka-dmytro (approved):

cxdong:
Thanks for caching this!

PR #118, opened 2026-09-02
Build fixes for pkvm-6.18

elmankku:
This PR adds two small fixes to address build failures when:

1. dynamic debug is disabled.
2. interrupt remapping is disabled.

### DIFF
--- a/arch/x86/include/asm/pkvm_image_vars.h
+++ b/arch/x86/include/asm/pkvm_image_vars.h
@@ -11,17 +11,27 @@
  */
 #define PKVM_ALIAS(sym)  pkvm_sym(sym) = sym;
 
+#ifdef CONFIG_PRINTK
 PKVM_ALIAS(_printk);
+PKVM_ALIAS(mem_dump_obj);
+#endif
+
 #ifdef CONFIG_BUG
 PKVM_ALIAS(__warn_printk);
 #endif
-PKVM_ALIAS(___ratelimit);
+
+#ifdef CONFIG_TRACING
 PKVM_ALIAS(__trace_bputs);
 PKVM_ALIAS(__trace_bprintk);
+#endif
+
+#ifdef CONFIG_DYNAMIC_DEBUG_CORE
 PKVM_ALIAS(__dynamic_pr_debug);
-PKVM_ALIAS(mem_dump_obj);
+#endif
+
+PKVM_ALIAS(___ratelimit);
 PKVM_ALIAS(vmalloc_base);
 PKVM_ALIAS(get_cpu_entry_area);
-#endif
+#endif /* CONFIG_PKVM_X86_DEBUG */
 
 #endif /* _ASM_X86_PKVM_IMAGE_VARS_H */
--- a/arch/x86/kvm/Kconfig
+++ b/arch/x86/kvm/Kconfig
@@ -137,7 +137,7 @@ config PKVM_X86
 config PKVM_INTEL
 	bool "pKVM for Intel processors support"
 	depends on KVM_INTEL=y
-	depends on INTEL_IOMMU
+	depends on INTEL_IOMMU && IRQ_REMAP
 	depends on BLK_DEV_FD=n
 	depends on X86_64
 	depends on X86_X2APIC


arch/x86/kvm/Kconfig
```diff
@@ -138,6 +138,7 @@ config PKVM_INTEL
 	bool "pKVM for Intel processors support"
 	depends on KVM_INTEL=y
 	depends on INTEL_IOMMU
+	depends on IRQ_REMAP
```
cxdong: Thanks for reporting this! It looks like the INTEL_IOMMU doesn't depends on IRQ_REMAP from the Kconfig. I suppose pKVM also doesn't need to depend on IRQ_REMAP, but seems like if we set CONFIG_IRQ_REMAP=n, pKVM compiling is failed:
```
../arch/x86/kvm/pkvm/../../../../drivers/iommu/intel/pkvm/iommu_hc.c:549:20: error: no member named 'irte' in 'struct irte'
  549 |                 u128 old = irte->irte;
      |                            ~~~~  ^
../arch/x86/kvm/pkvm/../../../../drivers/iommu/intel/pkvm/iommu_hc.c:557:33: error: no member named 'irte' in 'struct irte'
  557 |                 BUG_ON(!try_cmpxchg128(&irte->irte, &old, modified.irte));
      |                                         ~~~~  ^
../include/linux/atomic/atomic-instrumented.h:4951:9: note: expanded from macro 'try_cmpxchg128'
 4951 |         typeof(ptr) __ai_ptr = (ptr); \
      |                ^~~
../arch/x86/include/asm/pkvm_redef.h:72:45: note: expanded from macro 'BUG_ON'
   72 | #define BUG_ON(condition) do { if (unlikely(condition)) BUG(); } while (0)
      |                                             ^~~~~~~~~
../include/linux/compiler.h:77:42: note: expanded from macro 'unlikely'
   77 | # define unlikely(x)    __builtin_expect(!!(x), 0)
      |                                             ^
../arch/x86/kvm/pkvm/../../../../drivers/iommu/intel/pkvm/iommu_hc.c:557:33: error: no member named 'irte' in 'struct irte'
  557 |                 BUG_ON(!try_cmpxchg128(&irte->irte, &old, modified.irte));
      |                                         ~~~~  ^
../include/linux/atomic/atomic-instrumented.h:4951:26: note: expanded from macro 'try_cmpxchg128'
 4951 |         typeof(ptr) __ai_ptr = (ptr); \
      |                                 ^~~
../arch/x86/include/asm/pkvm_redef.h:72:45: note: expanded from macro 'BUG_ON'
   72 | #define BUG_ON(condition) do { if (unlikely(condition)) BUG(); } while (0)
      |                                             ^~~~~~~~~
../include/linux/compiler.h:77:42: note: expanded from macro 'unlikely'
   77 | # define unlikely(x)    __builtin_expect(!!(x), 0)
      |                                             ^
../arch/x86/kvm/pkvm/../../../../drivers/iommu/intel/pkvm/iommu_hc.c:557:54: error: no member named 'irte' in 'struct irte'
  557 |                 BUG_ON(!try_cmpxchg128(&irte->irte, &old, modified.irte));
      |                                                           ~~~~~~~~ ^
../include/linux/atomic/atomic-instrumented.h:4956:42: note: expanded from macro 'try_cmpxchg128'
 4956 |         raw_try_cmpxchg128(__ai_ptr, __ai_oldp, __VA_ARGS__); \
      |                                                 ^~~~~~~~~~~
../arch/x86/include/asm/pkvm_redef.h:72:45: note: expanded from macro 'BUG_ON'
   72 | #define BUG_ON(condition) do { if (unlikely(condition)) BUG(); } while (0)
      |                                             ^~~~~~~~~
../include/linux/compiler.h:77:42: note: expanded from macro 'unlikely'
   77 | # define unlikely(x)    __builtin_expect(!!(x), 0)
      |                                             ^
4 errors generated.
make[6]: *** [../arch/x86/kvm/pkvm/Makefile:96: arch/x86/kvm/pkvm/../../../../drivers/iommu/intel/pkvm/iommu_hc.pkvm.o] Error 1

```
It looks like the pKVM code should be fixed by properly hornoring the CONFIG_IRQ_REMAP rather than the Kconfig?

arch/x86/include/asm/pkvm_image_vars.h
```diff
@@ -18,7 +18,9 @@ PKVM_ALIAS(__warn_printk);
 PKVM_ALIAS(___ratelimit);
 PKVM_ALIAS(__trace_bputs);
 PKVM_ALIAS(__trace_bprintk);
```
maluka-dmytro, line 25: While we're at it, should we similarly put these two lines under `#ifdef CONFIG_TRACING`?

arch/x86/kvm/Kconfig
maluka-dmytro: Alternatively we could add `#ifdef CONFIG_IRQ_REMAP` in the pKVM code where needed (since in fact pKVM does not depend on the IRQ remapping functionality, I think), but it seems there is no real need for that, so this LGTM.

@vineethrp @aashishsharma7 for visibility.

arch/x86/include/asm/pkvm_image_vars.h
```diff
@@ -18,7 +18,9 @@ PKVM_ALIAS(__warn_printk);
...
+#ifdef CONFIG_DYNAMIC_DEBUG
 PKVM_ALIAS(__dynamic_pr_debug);
```
cxdong, line 29: __dynamic_pr_debug is enabled by CONFIG_DYNAMIC_DEBUG_CORE, and CONFIG_DYNAMIC_DEBUG selects CONFIG_DYNAMIC_DEBUG_CORE, so CONFIG_DYNAMIC_DEBUG would be N while CONFIG_DYNAMIC_DEBUG_CORE is Y, and in this case, __dynamic_pr_debug is there. So we can change the #ifdef CONFIG_DYNAMIC_DEBUG to #ifdef CONFIG_DYNAMIC_DEBUG_CORE?

maluka-dmytro: > so CONFIG_DYNAMIC_DEBUG would be N while CONFIG_DYNAMIC_DEBUG_CORE is Y, and in this case, __dynamic_pr_debug is there.

It is there, but do we need it in this case?

cxdong: Without CONFIG_DYNAMIC_DEBUG, __dynamic_pr_debug is not needed by pKVM. So if we treat #ifdef CONFIG_DYNAMIC_DEBUG as a gate to tell if the symbole is needed or not, I am fine with it.

Actually my first impression of using #ifdef CONFIG_DYNAMIC_DEBUG is to tell if the wrapped symbole is defined by this kconfig or not. From this point of view, it is a bit misleading as __dynamic_pr_debug is defined in lib/dynamic_debug.c and compiled by obj-$(CONFIG_DYNAMIC_DEBUG_CORE) += dynamic_debug.o. So the symbol is gated by CONFIG_DYNAMIC_DEBUG_CORE.

maluka-dmytro: Ok, I'm fine either way.

arch/x86/include/asm/pkvm_image_vars.h
elmankku: Sure, will do.

arch/x86/include/asm/pkvm_image_vars.h
elmankku: Okay, changing.

arch/x86/kvm/Kconfig
elmankku: Yep, I'll change that.

elmankku: Actually, this turned out to be trickier, because pKVM currently solely relies on x2APIC. That creates a dependency to interrupt remapping.

The dependency is mentioned in [this comment](https://github.com/intel-staging/pKVM-IA/blob/e58ec8928b95c07c4282a1526b16fd9f59039ada/drivers/iommu/intel/pkvm/iommu.c#L942) which is the reason why I initially thought `depends on IRQ_REMAP` to Kconfig.

I dug out the reason behind this and indeed, device interrupt routing on x2APIC mode depends on interrupt remapping. SDM Vol 3A `13.12.6 Routing of Device Interrupts in x2APIC Mode` states that

> This is made possible through use of the interrupt-remapping architecture specified in the Intel®
Virtualization Technology for Directed I/O Specification, Revision 1.3 and/or later versions, for the routing of interrupts from MSI-capable devices to local APIC units operating in x2APIC mode.

This is followed by `13.12.7 Initialization by System Software`:

> Routing of device interrupts to local APIC units operating in x2APIC mode requires use of the interrupt-remapping architecture specified in the Intel® Virtualization Technology for Directed I/O Specification (Revision 1.3 and/or later versions). Because of this, BIOS must enumerate support for and software must enable this interrupt remapping with Extended Interrupt Mode Enabled before it enabling x2APIC mode in the local APIC units.

Linux also [disables x2APIC mode](https://github.com/intel-staging/pKVM-IA/blob/e58ec8928b95c07c4282a1526b16fd9f59039ada/arch/x86/kernel/apic/apic.c#L1832) if interrupt remapping is not enabled. So first pKVM would need xAPIC support and then change IOMMU to handle CONFIG_IRQ_REMAP properly. Is this what we want?

Without xAPIC support, `depends on IRQ_REMAP` seems like right thing to do. Otherwise we allow compilation of unsupported configuration.

maluka-dmytro: > Linux also [disables x2APIC mode](https://github.com/intel-staging/pKVM-IA/blob/e58ec8928b95c07c4282a1526b16fd9f59039ada/arch/x86/kernel/apic/apic.c#L1832) if interrupt remapping is not enabled.

As that comment in Linux kernel says, it disables x2APIC if running on bare metal, but may still keep it enabled if running as a guest under a hypervisor whose `x86_init.hyper.x2apic_available()` returns true even with CONFIG_IRQ_REMAP=n. (If Linux always disabled x2APIC if IRQ remapping is not enabled, it could just make X2APIC depend on IRQ_REMAP at the Kconfig level, instead of runtime checks.)

That said, I suppose the use case of running pKVM inside a guest in such a setup that for some reason _requires_ IRQ remapping to be disabled is quite an exotic use case. (I mean, I run pKVM inside a QEMU guest under KVM very often, for development, but it's not a problem for me at all to keep CONFIG_IRQ_REMAP enabled in this guest's kernel.) So I'd probably be totally fine with keeping PKVM depending on IRQ_REMAP in the Kconfig.

maluka-dmytro: So, I didn't check how many #ifdefs we'd need to to add to pKVM code to make it compile with CONFIG_IRQ_REMAP=n (and work correctly with CONFIG_IRQ_REMAP=n). If that's too many #ifdefs or it's too ugly, then it's probably better to go with `depends on IRQ_REMAP` in Kconfig for simplicity. If it's just a couple of non-invasive #ifdefs, then maybe worth adding those #ifdefs instead.

elmankku: I suppose not too many ifdefs, but to making it correct would complicate pKVM's IOMMU side. Not sure if it would be worth it.

Your call.

elmankku:
Updated patches to use CONFIG_DYNAMIC_DEBUG_CORE instead of CONFIG_DYNAMIC_DEBUG. Also, now handling CONFIG_PRINTK and CONFIG_TRACING. Tested with/without each config.

Keeping `depends on IRQ_REMAP` until discussion resolved.

arch/x86/include/asm/pkvm_image_vars.h
```diff
@@ -11,17 +11,27 @@
  */
 #define PKVM_ALIAS(sym)  pkvm_sym(sym) = sym;
 
+#ifdef CONFIG_PRINTK
```
maluka-dmytro, line 14: There are no-op versions of `_printk()` and `mem_dump_obj()` available if CONFIG_PRINTK is disabled, so this is not necessary?

elmankku: Yes, it is required. Without ifdef and CONFIG_PRINTK we get:
```
/nix/store/wcv8n2k53w8sbffpf716gxj6mjb5jf26-binutils-2.44/bin/ld:./arch/x86/kernel/vmlinux.lds:201: undefined symbol `_printk' referenced in expression
```
And the same applies to `mem_dump_obj()`.

maluka-dmytro: Ah ok, I guess `_printk` is not a symbol then (since it is defined as a static inline function, and not referenced by pointer anywhere in the code, only in the linker script).

arch/x86/include/asm/pkvm_image_vars.h
```diff
@@ -11,17 +11,27 @@
  */
```
maluka-dmytro, line 11: Worth adding `Fixes: 7f844b104302 ("DEBUG: pKVM: x86: Add log printing capability for debug")` to the commit message?

elmankku: Yes, done. :+1:

maluka-dmytro: nit: unneeded/unconventional empty line between `Fixes:` and `Signed-off-by:`? :)

arch/x86/kvm/Kconfig
cxdong: yeah, I also think there should be no too many ifdefs according to the compiling failure. Just a few to wrap pkvm_iommu_modify_irte(), or maybe pkvm_modify_irte() as well.

If we want to add CONFIG_IRQ_REMAP dependency to pKVM due to its the X86_X2APIC's dependency, is there more dependencies need to be added? Actually the reason we want to add this dependency is just because of the compiling depdency.

CONFIG_X86_X2APIC=y doesn't guarantee the x2apic mode is always enabled, like the case CONFIG_IRQ_REMAP=n && CONFIG_HYPERVISOR_GUEST=y, CONFIG_X86_X2APIC will be y but this kernel is failed to enable x2apic on native. So similar to pKVM, with CONFIG_IRQ_REMAP=n && CONFIG_X86_X2APIC=y, CONFIG_PKVM_INTEL will be y but this kernel is falied to initialize pKVM on native. So is this acceptable?

If we all accept this, then we don't need to add depedency with CONFIG_IRQ_REMAP to pKVM but just add #ifdef CONFIG_IRQ_REMAP to wrap pkvm_iommu_modify_irte(), or maybe pkvm_modify_irte() as well.

arch/x86/include/asm/pkvm_image_vars.h
elmankku: Yep, that stands out :) Removing...

arch/x86/kvm/Kconfig
elmankku: While `CONFIG_X86_X2APIC=y` does not guarantee that the x2apic mode will be enabled, the configuration makes it possible for pKVM to work properly. The difference to `CONFIG_IRQ_REMAP=n` is that in its current shape, pKVM would always fail [here](https://github.com/intel-staging/pKVM-IA/blob/e58ec8928b95c07c4282a1526b16fd9f59039ada/drivers/iommu/intel/pkvm/iommu.c#L944). So without this patch, we would allow building an unsupported configuration that is only caught during runtime. And that is different. The check itself is needed for correctness.

Although we would fix the compilation errors, `CONFIG_IRQ_REMAP=n && CONFIG_HYPERVISOR_GUEST=y` would be a different operating mode that should be taken into account at least in host IOMMU capability checks, IOMMU initialization and MMIO handling. So it would add some complexity.

My recommendation would be either to apply this patch or alternatively add support for at least one CONFIG_IRQ_REMAP=n configuration (native or guest).

maluka-dmytro: Looks like in order to correctly support pKVM with IRQ_REMAP=n, we'd need at least to: remove that `if (!(iommu->vgsts & DMA_GSTS_IRTPS) || !(iommu->vgsts & DMA_GSTS_IRES))` check preventing pKVM initialization if IR is not set up; but also avoid calling `iommu_protect_ir_table()` if DMA_GSTS_IRTPS is not set (because in this case, the IRTA value is not expected to be an actual address of an IR table prepared by the host).

But I've also recalled that we, in particular, require the IOMMU hardware to support IR, since commit 6a26a33df623 ("iommu/vt-d: pKVM: Enforce EIM and IR supported in ECAP"). So, shall we still require it even though we allow CONFIG_IRQ_REMAP=n? Or shall we rather allow ECAP.IR=0?

We are actually relying on ECAP.EIM=1, since it allows us to simplify various sanity checks when handling IOMMU MMIO accesses. _Maybe_ we are not strictly relying on ECAP.IR=1, but it might be a bit odd to require EIM without requiring IR?

And who knows how many other subtleties there are...

...Now I tend to think it doesn't make a lot of sense to try to support pKVM with IRQ_REMAP=n. As I understand, the reason why Linux allows x2APIC without IR in a guest is that guests often run without any virtualized IOMMU at all, while it is still useful to have x2APIC in a guest for various reasons. But in the pKVM case, pKVM requires an IOMMU anyway (e.g. pKVM initialization fails if there is no DMAR table, and so on), even in a guest.

So I tend to agree we should rather just require IRQ_REMAP=y, instead of complicating things without a real need.

After all, pKVM relies on architecturally defined IOMMU (VT-d) behavior, and architecturally IR is required to be enabled if x2APIC is enabled. (Even if pKVM doesn't rely on this particular aspect of the architectural behavior, we have no need to support a deviation from this aspect either?)

maluka-dmytro: ...Nit suggestion: maybe `depends on INTEL_IOMMU && IRQ_REMAP` instead of adding `depends on IRQ_REMAP` on a separate line (to keep it compact, and to emphasize that the two dependencies are related)?

vineethrp: > After all, pKVM relies on architecturally defined IOMMU (VT-d) behavior, and architecturally IR is required to be enabled if x2APIC is enabled. (Even if pKVM doesn't rely on this particular aspect of the architectural behavior, we have no need to support a deviation from this aspect either?)

> So I tend to agree we should rather just require IRQ_REMAP=y, instead of complicating things without a real need.

I also agree. Lets keep IRQ_REMAP=y to keep it simple.

cxdong: I see. Thanks for those discussion. Now I also agree with adding IRQ_REMAP depedency, given that pKVM requires IOMMU with IR caps in the code even running in the VM.

And maybe worth to use `depends on INTEL_IOMMU && IRQ_REMAP`  as Dmytro suggested?

elmankku: Yes, of course. :+1:  Done.

elmankku:
Updated, all discussions resolved.

maluka-dmytro (approved):

cxdong:
Thanks for the fix. Merge the patches.

